PR #23, opened 2018-11-09
small tweaks needed for using crds

clux:


shipcat_definitions/src/structs/metadata.rs
```diff
@@ -168,15 +168,15 @@ mod tests {
 
     #[test]
     fn valid_slack_channel() {
-        let sc = SlackChannel::new("#dev-platform".into());
```
JoseAlban: interesting :) thx

PR #25, opened 2018-11-12
minimal shipcat api skeleton

clux:
[8MB](https://quay.io/repository/babylonhealth/raftcat?tab=tags) api service that proxies CRD data from kube. Reuses the manifest struct from shipcat and parses straight into it.


Dockerfile.raftcat
```diff
@@ -0,0 +1,5 @@
+FROM alpine:latest
```
florent-babylon: The `:latest` is redundant here, isn't it? When omitted docker should go for the `latest` tag by default iirc.

Dockerfile.raftcat
```diff
@@ -0,0 +1,5 @@
...
+RUN apk --no-cache add ca-certificates
+ADD raftcat.x86_64-unknown-linux-musl /bin/raftcat
+EXPOSE 8080
+ENTRYPOINT ["/bin/raftcat"]
```
florent-babylon: `ENTRYPOINT` rather than `CMD`? Tbh I'm not sure what the difference is in case of a single executable. 🤷‍♂️

Makefile
```diff
@@ -87,4 +87,14 @@ kongfig-tag-latest:
 	docker tag  $(REPO)/$(NAME):kongfig-$(VERSION) $(REPO)/$(NAME):kongfig
 	docker push $(REPO)/$(NAME):kongfig
 
-.PHONY: doc install build compile releases
+# raftcat experiment
+raftcat:
+	docker run \
+		-v cargo-cache:/root/.cargo \
+		-v "$$PWD:/volume" -w /volume \
+		--rm -it clux/muslrust:stable cargo build --release --bin raftcat
```
florent-babylon: building from your own rust image? ;)

clux: it's one of the main ways to build musl images actually :slightly_smiling_face: 
It's got 1M pulls!

Dockerfile.raftcat
krubot: `CMD ["/bin/raftcat"]` translates to `/bin/sh -c /bin/raftcat` but `ENTRYPOINT ["/bin/raftcat"]` translates to `/bin/raftcat`. Less code is usually better 😄

Dockerfile.raftcat
clux: i'll pin it eventually. at least it's being explicit about doing a bad thing.

PR #33, opened 2018-11-16
Allow environment variables to have sidecars

### LINKED ISSUE #31: Set environment variables on sidecars
It would be useful if it was possible to set environment variables on sidecars from a service's manifest.

The values should be processed the same way as the main container's environment variables.

```yaml
sidecars:
- name: example-sidecar
  env:
    # Static value
    USERNAME: a-user
    # Value from Vault
    PASSWORD: IN_VAULT
    # Value from tera template
    URL: "{{ base_urls.services }}/oauth2/token"
```

janslow:
Fixes #31.

* [x] Support secrets in vault


shipcat_definitions/src/template.rs
```diff
@@ -136,8 +144,9 @@ impl Manifest {
     /// Template evars - must happen before inline templates!
     pub fn template_evars(&mut self, reg: &Region) -> Result<()> {
         let ctx = self.make_template_context(reg)?;
-        for (_, v) in &mut self.env {
-            *v = one_off(v, &ctx)?;
+        Manifest::template_env(&ctx, &mut self.env)?;
+        for s in &mut self.sidecars {
+            Manifest::template_env(&ctx, &mut s.env)?;
```
clux: Nice! Could we do the same for self.workers and self.cronjobs  while we are at it?

tests/services/fake-ask/shipcat.yml
```diff
@@ -22,6 +22,10 @@ metadata:
   gitTagTemplate: "prefix-{{ version }}-suffix"
 sidecars:
 - name: redis
+  env:
+    STATIC_VALUE: "static"
+    FAKE_SECRET: "IN_VAULT"
+    CORE_URL: "{{ base_urls.services }}/somesvc"
```
clux: :+1:

shipcat_definitions/src/structs/sidecar.rs
```diff
@@ -10,4 +12,37 @@ pub struct Sidecar {
...
+  ///
+  /// The vault lookup will GET from the region specific path for vault, in the
+  /// webapp subfolder, getting the `DATABASE_URL` secret.
+  ///
+  /// The `kong` templating will use the secrets read from the `Config` for this
+  /// region, and replace them internally.
+  ///
+  /// The `as_secret` destinction only serves to put `AUTH_SECRET` into `Manifest::secrets`.
+  #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+  pub env: BTreeMap<String, String>,
```
clux: env should probably use `EnvVar` which is what workers and cronjobs use. will eventually move the main env to use it as well.

clux: That way you don't have to duplicate all that doc for all users as well.

shipcat_definitions/src/template.rs
janslow: Done

shipcat_definitions/src/structs/sidecar.rs
janslow: Done

shipcat_definitions/src/structs/sidecar.rs
```diff
@@ -10,4 +12,14 @@ pub struct Sidecar {
 
   #[serde(skip_serializing_if = "Option::is_none")]
   pub resources: Option<Resources<String>>,
+
+  #[serde(default, skip_serializing_if = "EnvVars::is_empty")]
+  pub env: EnvVars,
+}
```
janslow: @clux, at the moment, it's not trivial to template the `IN_VAULT` templates, as the `Manifest.secret` field is used to generate the K8s secrets and the environment variables on the main container.

To support this for Sidecars etc., we'd either need to:
* Add a `secrets` field to `Sidecar`/`Worker`/`CronJob` as well, then add templating in the base charts to generate the K8s Secrets from each of these sources; or
* Split the `Manifest.secret` field into two, one for generating secrets and one which contains the names of environment variables to read from secrets (which would also be added to `Sidecar` etc); or
* Make `env` values a more complex object, so the value could either be a plain value or a `valueFrom`.

Any preferences? Or alternatives?

clux: Ah yes. I think option 1 and 3 could be combined here by:

- turning the `EnvVar` struct was turned into a more complicated object with a `secrets` output field on it
- exposing the union of all secrets in the k8s secrets object
- mounting the secrets relevant for the resource based on the output of `$.env.secret` from the unioned secrets object
- addind a validate rule to ensure secret names are unique across resources (sanity if doing one secret object)

This should be doeable **without** changing all the manifests to have a more complicated input API.

I feel that is likely to be more encapsulated than option 2. Both are fine really, but if you do go for an option 2. Then you can also combine the names and secrets in a single struct so all this special case logic is encapsulated, and the single output can be used in the chart to demux.

janslow: Now done (see PR in `manifests` for a PR to update our charts).

shipcat_cli/tests/common.rs
```diff
@@ -3,13 +3,26 @@ extern crate shipcat;
...
+    ( $( $x:expr ),* ) => {
+        {
+            let mut set = BTreeSet::new();
+            $(
+                set.insert($x);
+            )*
+            set
+        }
+    };
+}
```
clux: this ought to be in the standard library tbh :+1:

shipcat_definitions/src/structs/env.rs
```diff
@@ -1,19 +1,69 @@
...
+        value == "IN_VAULT"
+    }
 
+    fn template_secret_value(value: &String) -> Option<String> {
+        let prefix = "SHIPCAT_SECRET::";
+        if value.starts_with(prefix) {
+            Some(value.to_string().split_off(prefix.len()))
+        } else {
+            None
+        }
```
clux: nice functions :+1:

shipcat_definitions/src/structs/env.rs
```diff
@@ -1,19 +1,69 @@
...
+/// The `as_secret` destinction only serves to put `AUTH_SECRET` into `Manifest::secrets`.
+#[derive(Serialize, Clone, Default)]
+pub struct EnvVars {
+    /// Plain text (non-secret) environment variables
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub plain: BTreeMap<String, String>,
+
+    /// Environment variable names stored in secrets
+    ///
+    /// This is an internal property that is exposed as an output only.
```
clux: Good docs :+1:

shipcat_definitions/src/structs/env.rs
```diff
@@ -23,68 +73,51 @@ impl EnvVars {
...
+    where
+        D: Deserializer<'de>,
+    {
+        let plain = BTreeMap::deserialize(deserializer)?;
+        let secrets = BTreeSet::new();
 
-impl DerefMut for EnvVars {
-    fn deref_mut(&mut self) -> &mut BTreeMap<String, String> {
-        &mut self.0
+        Ok(EnvVars { plain, secrets })
```
clux: nice. so you start out in the pre-moved state. this appears to be the base/crd manifest state now.

shipcat_definitions/src/structs/env.rs
```diff
@@ -1,19 +1,69 @@
...
+        }
+    }
 
-impl EnvVars {
     pub fn verify(&self) -> Result<()> {
-        for (k, v) in *&self {
-            if v == "IN_VAULT" {
+        for (k, v) in &self.plain {
+            if EnvVars::is_vault_secret(v) {
                 // TODO: can do this generally here now
```
clux: this TODO no longer applies, right?

shipcat_definitions/src/structs/worker.rs
```diff
@@ -33,7 +33,7 @@ pub struct Worker {
     ///
     /// These may be specified in addition to the main deployment `env` vars
     /// or as fresh variables, depending on `preserveEnv`.
-    #[serde(default, skip_serializing_if = "EnvVars::is_empty")]
+    #[serde(default)]
```
clux: I guess you are dealing with a potentially empty vector returned to the charts, in the charts.

shipcat_definitions/src/structs/env.rs
```diff
@@ -23,68 +73,51 @@ impl EnvVars {
...
+        for (k, v) in self.plain.iter() {
+            match EnvVars::template_secret_value(v) {
+                Some(x) => {
+                    ts.insert(k.to_string(), x);
+                    self.secrets.insert(k.to_string());
+                }
+                None => {
+                    plain.insert(k.to_string(), v.to_string());
+                }
+            };
```
clux: very nice function :+1: 
sidenote: this is maybe an `.iter()` with a `.partition`, but that probably doesn't fit in here.

shipcat_definitions/src/manifest.rs
```diff
@@ -867,23 +882,34 @@ impl Manifest {
...
-            if v == "IN_VAULT" {
-                let vkey = format!("{}/{}", pth, k);
-                self.secrets.insert(k.to_string(), client.read(&vkey)?);
+        let mut vault_secrets = BTreeSet::new();
+        let mut template_secrets = BTreeMap::new();
+        for e in &mut self.get_env_vars() {
+            for k in &mut e.vault_secrets().iter() {
+                if template_secrets.contains_key(k) {
+                    bail!("Secret {} can not be templated and fetched from vault", k);
+                }
```
clux: i feel this bail needs a `both` as an operative word here. took me a while to get what this was doing.

shipcat_definitions/src/manifest.rs
```diff
@@ -867,23 +882,34 @@ impl Manifest {
...
-        let special = "SHIPCAT_SECRET::".to_string();
-        // iterate over key value evars and replace placeholders
-        for (k, v) in &mut self.env {
-            if v == "IN_VAULT" {
-                let vkey = format!("{}/{}", pth, k);
-                self.secrets.insert(k.to_string(), client.read(&vkey)?);
+        let mut vault_secrets = BTreeSet::new();
+        let mut template_secrets = BTreeMap::new();
+        for e in &mut self.get_env_vars() {
+            for k in &mut e.vault_secrets().iter() {
```
clux: `&mut e.vault_secrets().iter()` wtf?
also, you're not mutating k or e here?

is this a consequence of `get_env_vars` returning mut refs?

shipcat_definitions/src/manifest.rs
```diff
@@ -867,23 +882,34 @@ impl Manifest {
...
         debug!("Injecting secrets from vault {} ({:?})", pth, client.mode());
 
-        let special = "SHIPCAT_SECRET::".to_string();
-        // iterate over key value evars and replace placeholders
-        for (k, v) in &mut self.env {
-            if v == "IN_VAULT" {
-                let vkey = format!("{}/{}", pth, k);
-                self.secrets.insert(k.to_string(), client.read(&vkey)?);
+        let mut vault_secrets = BTreeSet::new();
+        let mut template_secrets = BTreeMap::new();
```
clux: since these are just sets, you can just verify after that their intersection is empty. easier to read.

clux:
Also, this needs a minor bump :-)

shipcat_cli/tests/common.rs
janslow: Yeah, especially as they already have `vec!`

shipcat_definitions/src/structs/worker.rs
janslow: Yep, otherwise you'd have to always guard `.Values.env.plain` etc. The `plain` and `secrets` properties are skipped if empty (as `range nil` is safe in go).

shipcat_definitions/src/structs/env.rs
janslow: Done

shipcat_definitions/src/manifest.rs
janslow: Done

shipcat_definitions/src/manifest.rs
janslow: Now massively simplified.

shipcat_definitions/src/manifest.rs
```diff
@@ -867,23 +882,33 @@ impl Manifest {
...
+                vault_secrets.insert(k.to_string());
             }
-            // Special cases that were handled by `| as_secret` template fn
-            if v.starts_with(&special) {
-                self.secrets.insert(k.to_string(), v.split_off(special.len()));
+            for (k, v) in e.template_secrets() {
+                let original = template_secrets.insert(k.to_string(), v.to_string());
+                if original.is_some() {
+                    bail!("Secret {} can not be used in two templates", k);
+                }
```
clux: interesting constraint, where does that come from?

janslow: It's designed to prevent the following:

```yaml
env:
  FOO: "{{ 1 | as_secret }}"
sidecars:
  - env:
      FOO: "{{ 2 | as_secret }}"
```

i.e., `FOO`'s secret value will be `2`, so the main container's env will be wrong.

However, I can change this to permit reuse if the values are the same.

janslow: Done

janslow:
Just realised I've introduced an unused variable warning, will merge after I've fixed that

PR #34, opened 2018-11-19
Make raftcat do a whole lot of things

clux:
Nice service dashboard on /service/{service}

Finds out a lot of implicit info about where go.

Reads a lot on boot:

- manifest crds
- config crd
- version info (from version endpoint) if present
- newrelic info (from newrelic) if evars present
- sentry slugs (from sentry) if evars present


raftcat/src/main.rs
```diff
@@ -1,80 +1,136 @@
...
+    pub relics: RelicMap,
+    pub sentries: SentryMap,
     last_update: Instant,
 }
 impl AppState {
     pub fn new(client: APIClient) -> Self {
         info!("Loading state from CRDs");
-        let manifests = kube::get_shipcat_manifests(&client).unwrap();
-        AppState {
+        let config = kube::get_shipcat_config(&client, "dev-uk").unwrap().spec;
```
florent-babylon: Hardcoded to dev-uk? 🤔

clux: ah, i need to pick up on SERVICE_REGION

PR #39, opened 2018-11-22
Make raftcat look pretty (for services at least)

MentalAtom:
image


raftcat/templates/service.tera
```diff
@@ -119,6 +119,7 @@
                         <td>
                           {% if value == "IN_VAULT" %}
                             <a href="{{ vault_link }}/{{key}}">{{value}}</a>
+                            <!-- TODO: baseurls templating  -->
```
MentalAtom: @clux I was gonna ask about this :D

clux: it's non-trivial i believe

PR #42, opened 2018-11-27
Move from error-chain to failure

clux:
This  was a bit of a yak that needed to be shaved to see if it was helping my `!Sync` serde errors. But don't think I actually need this. Yet. It's certainly pretty awkward in the current form, although it does have more potential than error-chain.

Stashing it here for now. Comments welcome, but this is likely going to linger for a while until `failure` stabilizes.


clux:
Might wanna drop this and use [failchain](https://github.com/cristicbz/failchain) for the cli.

PR #48, opened 2018-12-04
Audit upgrade and reconcile actions

JoseAlban:
Just like telling slack, we also tell a configured audit backend that an action is about to happen, and has happened.
Actions ATM being: helm upgrade / reconcile.

- Warn fast if auditing but config missing
- Audit high-level reconcile and individual upgrades/deployment
- Provide generic framework for more webhooks in future - and defer slack/grafana webhook refactoring


shipcat_cli/src/helm/direct.rs
```diff
@@ -68,6 +69,39 @@ impl UpgradeMode {
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            &UpgradeState::Pending => write!(f, "PENDING"),
+            &UpgradeState::Completed => write!(f, "COMPLETED"),
+            &UpgradeState::Failed => write!(f, "FAILED"),
+            &UpgradeState::RollingBack => write!(f, "ROLLING_BACK"),
+            &UpgradeState::RolledBack => write!(f, "ROLLED_BACK"),
+        }
+    }
+}
```
clux: Just derive `Debug`? This is almost equivalent to that.

shipcat_definitions/src/region.rs
```diff
@@ -82,10 +84,29 @@ pub struct KafkaConfig {
...
+    /// Audit webhook details
+    pub audit: AuditWebhook,
+}
+
+/// Where / how to send audited events
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct AuditWebhook {
+    /// Endpoint
+    #[serde(with = "url_serde")]
```
clux: we shouldn't need this extra deprecated crate. just use url with the serde feature:
https://docs.rs/url/1.7.2/src/url/lib.rs.html#2248

shipcat_definitions/src/region.rs
```diff
@@ -204,7 +225,7 @@ impl KongConfig {
     }
 }
 
-#[derive(Serialize, Deserialize, Clone, Default)]
+#[derive(Serialize, Deserialize, Clone, Debug, Default)]
```
clux: why all these debug impls?

shipcat_definitions/src/region.rs
```diff
@@ -218,13 +239,34 @@ impl KongConfig {
     }
 }
 
+impl Webhooks {
+    fn secrets(&mut self, vault: &Vault, region: &str) -> Result<()> {
+        let webhook_audit = &mut self.audit;
+        if webhook_audit.token == "IN_VAULT" {
+            let vkey = format!("{}/shipcat/WEBHOOK_AUDIT_TOKEN", region);
```
clux: ah shipcat scoped secrets, makes sense

shipcat_definitions/src/region.rs
```diff
@@ -267,13 +309,18 @@ pub struct Region {
     /// List of locations the region serves
     #[serde(default)]
     pub locations: Vec<String>,
+    /// All webhooks
+    pub webhooks: Option<Webhooks>,
```
clux: I still feel this should be a `Vec<Webhook>`

shipcat_definitions/src/region.rs
```diff
@@ -82,10 +84,29 @@ pub struct KafkaConfig {
...
+/// Where / how to send audited events
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct AuditWebhook {
+    /// Endpoint
+    #[serde(with = "url_serde")]
+    pub url: Url,
+    /// Credential
+    pub token: String,
+}
```
clux: If these are all the fields, then we can just have a Webhook<T> and put a `payload: T` in the body

shipcat_cli/src/audit.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// Timestamped payload skeleton
+    pub fn new(dt: &str, us: &UpgradeState) -> Self {
+        AuditDeploymentPayload{
+            domainType: String::from(dt),
+            timestamp: format!("{}", Utc::now().to_rfc3339_opts(SecondsFormat::Millis, true)),
+            status: us.clone(),
+            upstreamId: match env::var("SHIPCAT_AUDIT_UPSTREAM_ID") {
+                Ok(ev) => Some(ev),
+                _ => None,
+            },
```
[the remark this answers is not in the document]
clux: oh actually, it's an [.ok()](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok)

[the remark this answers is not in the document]
JoseAlban: thanks! that's what I was looking for :)

shipcat_cli/src/audit.rs
```diff
@@ -0,0 +1,118 @@
...
+            ("unknown".into(), "unknown".into(), "unknown".into())
+        };
+        let rid = "todo".into();
+
+        AuditDeployment{
+            id: format!("{}-{}-{}", service, version, rid),
+            region: region,
+            revisionId: rid,
+            service: service,
+            version: version,
```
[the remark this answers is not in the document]
JoseAlban: oh, didn't know about that :)

shipcat_cli/src/audit.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub id: String,
+    pub region: String,
+    /// Eg Git SHA
+    pub revisionId: String,
+    pub service: String,
+    pub version: String,
+}
+
+impl AuditDeployment {
+    pub fn new(udopt: &Option<UpgradeData>, ) -> Self {
```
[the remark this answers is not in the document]
JoseAlban: I was concerned it's not idiomatic indeed, but not clear on the reasoning yet :/ - will (un)do

shipcat_definitions/src/region.rs
JoseAlban: So that we can print the object during tests with `"{:?}"`

shipcat_definitions/src/region.rs
JoseAlban: yes, a bit bruteforce having it hardcoded but.. i can imagine more usecases.

shipcat_definitions/src/region.rs
JoseAlban: Hmm, Vec means we need to iterate/find key, struct == key directly :|

shipcat_definitions/src/region.rs
JoseAlban: could do - that'd be a step forward generalising the concept - I'd call it pre-optimisation at this point though.

shipcat_cli/src/audit.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// Domain Object
+    pub deployment: AuditDeployment,
+}
+
+impl AuditDeploymentPayload {
+    /// Timestamped payload skeleton
+    pub fn new(dt: &str, us: &UpgradeState) -> Self {
+        AuditDeploymentPayload{
+            domainType: String::from(dt),
+            timestamp: format!("{}", Utc::now().to_rfc3339_opts(SecondsFormat::Millis, true)),
```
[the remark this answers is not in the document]
clux: ah cool. You don't need to format the string to get a string btw. it's already a string.

shipcat_definitions/src/region.rs
clux: `#[cfg_attr(test, derive(Debug))]` is a thing

shipcat_cli/src/audit.rs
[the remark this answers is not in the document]
JoseAlban: good point - err'ing on the safe side too much

shipcat_definitions/src/region.rs
JoseAlban: huh good spot - although I can not find references for the deprecation around (no warning from CLIs / online docs). Also, I failed when trying to:
`url = { version = "1.7.2", features = ["serde"] }`
and removing the serde attribute for the field. Compiler wasn't happy (and I don't understand why, as it can't find serde's trait impls for Url)
"the trait `serde::Serialize` is not implemented for `url::Url`"

So, will leave `url_serde` for now.

shipcat_cli/src/helm/direct.rs
JoseAlban:
```
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn upgrade_state_debug_string() {
        assert_eq!(format!("{:?}", UpgradeState::RolledBack), "ROLLED_BACK");
    }
}
```

```
 left: `"RolledBack"`,
 right: `"ROLLED_BACK"`', shipcat_cli/src/helm/direct.rs:99:9
```

It'd be equivalent if I could call like `UpgradeState::RolledBack.serialize()` for the right case :/

clux: `serde_json::to_string_pretty` will invoke serialize

shipcat_cli/src/audit.rs
```diff
@@ -8,108 +8,135 @@ use super::AuditWebhook;
...
 
 /// Payload that gets sent via audit webhook
-#[derive(Serialize, Clone, Default)]
+#[derive(Serialize, Clone)]
 #[cfg_attr(test, derive(Debug))]
-#[serde(rename_all = "snake_case")]
-struct AuditDeploymentPayload {
-    #[serde(rename = "type")]
-    pub domainType: String,
+// #[serde(rename_all = "snake_case")] // well, it just didn't work. :/ maybe on a later serde version?
```
clux: maybe it interacts badly with flatten

shipcat_cli/src/audit.rs
```diff
@@ -8,108 +8,135 @@ use super::AuditWebhook;
...
     pub contextId: Option<String>,
     /// Eg a jenkins job url
-    #[serde(with = "url_serde")]
+    #[serde(rename = "context_link", with = "url_serde", skip_serializing_if = "Option::is_none")]
     pub contextLink: Option<Url>,
-    /// Domain Object
-    pub deployment: AuditDeployment,
+
+    /// represents a single helm upgrade or a reconciliation
+    #[serde(flatten)]
```
clux: are you sure you want to flatten this?

shipcat_cli/src/audit.rs
```diff
@@ -8,108 +8,135 @@ use super::AuditWebhook;
...
-        AuditDeploymentPayload{
-            domainType: String::from(dt),
+    pub fn new(us: &UpgradeState) -> Self {
+        AuditEvent{
             timestamp: Utc::now().to_rfc3339_opts(SecondsFormat::Millis, true),
             status: us.clone(),
             contextId: env::var("SHIPCAT_AUDIT_CONTEXT_ID").ok(),
             contextLink: if let Some(urlstr) = env::var("SHIPCAT_AUDIT_CONTEXT_LINK").ok() {
                 url::Url::parse(&urlstr).ok()
             } else { None },
```
clux: `env::var("..").map(Url::parse).map(Result::ok)`

shipcat_cli/src/helm/direct.rs
```diff
@@ -517,7 +520,12 @@ pub fn handle_upgrade_notifies(us: &UpgradeState, ud: &UpgradeData, r: &Region)
         }
         _ => {
             if let Some(ref webhooks) = &r.webhooks {
-                audit(&us, &ud, &webhooks.audit)
+                if let Err(e) = audit_deployment(&us, &ud, &webhooks.audit) {
+                    warn!("Failed to notify about deployment: {}", e);
+                    Err(e)
```
clux: this is a chain_err

shipcat_cli/src/helm/parallel.rs
```diff
@@ -49,16 +55,29 @@ pub fn reconcile(svcs: Vec<Manifest>, conf: &Config, region: &Region, umode: Upg
...
         match e {
             Error(ErrorKind::MissingRollingVersion(svc),_) => {
                 // This only happens in rolling envs because version is mandatory in other envs
                 warn!("'{}' missing version for {} - please add or install", svc, region.name);
             },
             // remaining cases not ignorable
-            e => return Err(e),
+            e => {
+                anyNonIgnorableError = true;
+                return Err(e)
```
clux: this bool is always true if !res.is_empty() further down - it's a bit awward

shipcat_cli/src/audit.rs
JoseAlban: Yes TBH, as that was the cleanest way to get it to match the API we've designed for events - so we can use `type` and `payload` for the enum tag/content and that matches the spec just right :)

JoseAlban: The alternative would be to have `type` back in and `payload` a generic type, which might also work but then there's a generic type which is arguably more complex than this.

shipcat_cli/src/audit.rs
JoseAlban: Excellent!

shipcat_cli/src/helm/parallel.rs
JoseAlban: I actually ended up removing it when I realised the logic was wrong.

shipcat_cli/src/audit.rs
JoseAlban: or actually.. not that straight-forward, type mismatch all the way :/

shipcat_cli/src/audit.rs
JoseAlban: Hmm tried that assumption but no avail - commenting out `flatten` didn't make it work. I'll just remove the attribute at struct level as well, it was more for ref on this PR that i've at least tried.

shipcat_cli/src/helm/direct.rs
JoseAlban: It'd be indeed - but then I realised we don't wanna panic if failed to notify audit backend, so changed it to be always `Ok(())`

JoseAlban:
`shipcat_test` is probably failing since a few days because the webhook integration has been removed ;(

JoseAlban:
I reckon we should have a circleci's `cargo_test_optional` for tests like `slack_test` which has flaky dependencies eg webhook must be up.

shipcat_definitions/Cargo.toml
```diff
@@ -4,12 +4,13 @@ authors = [
   "Florent Delannoy <florent.delannoy@babylonhealth.com>",
   "Jeremie Vallee <jeremie.vallee@babylonhealth.com>",
   "Laurent Godet <laurent.godet@babylonhealth.com>",
-  "Guy Leaver <guy.leaver@babylonhealth.com>"
+  "Guy Leaver <guy.leaver@babylonhealth.com>",
+  "Jose Alban <jose.alban@babylonhealth.com>"
```
clux: comma at end

shipcat_definitions/src/config.rs
```diff
@@ -91,6 +96,7 @@ pub struct GithubParameters {
 
 
 #[derive(Serialize, Deserialize, Clone)]
+#[cfg_attr(test, derive(Debug))]
```
clux: i thought we were taking these out?

JoseAlban: oh yeah - forgot to remove after using again. will rm

shipcat_cli/src/helm/direct.rs
```diff
@@ -489,6 +441,8 @@ pub fn values_wrapper(svc: &str, region: &Region, conf: &Config, ver: Option<Str
 
 /// Full helm wrapper for a single upgrade/diff/install
 pub fn upgrade_wrapper(svc: &str, mode: UpgradeMode, region: &Region, conf: &Config, ver: Option<String>) -> Result<Option<UpgradeData>> {
+    webhooks::ensure_requirements(&region)?;
```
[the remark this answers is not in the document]
JoseAlban: it will only break if they've set up audit webhook though, which is the expected behaviour, right?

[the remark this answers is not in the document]
JoseAlban: otherwise we wouldn't even need to ensure requirements - or warn maybe.

shipcat_definitions/src/config.rs
JoseAlban: reverted all Debug derivations

shipcat_cli/src/webhooks.rs
```diff
@@ -0,0 +1,182 @@
...
+
+/// Throw events to configured webhooks - warning on delivery errors
+///
+/// Http errors are NOT propagated from here
+pub fn reconcile_event(state: UpgradeState, reg: &Region) {
+    if let Some(whs) = &reg.webhooks {
+        for wh in whs.iter() {
+            if let Err(e) = match wh {
+                Webhook::Audit(h) => {
+                    audit::audit_reconciliation(&state, &reg.name, &h)
```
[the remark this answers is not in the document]
JoseAlban: that's a good point - we'll just lose the ability to fail fast though :/ eg a reconcile will gather info first before failing when trying to send the first audit event.

shipcat_cli/src/webhooks.rs
```diff
@@ -0,0 +1,177 @@
...
+            }
+        }
+    }
+
+    // Slack and Grafana
+
+    let code = if ud.diff.is_empty() { None } else { Some(ud.diff.clone()) };
+    let (color, text) = match us {
+        UpgradeState::Completed => {
+            info!("successfully rolled out {}", ud.name);
```
clux: move these to calling place instead!

shipcat_cli/src/webhooks.rs
```diff
@@ -0,0 +1,177 @@
...
+            ("good".into(), format!("{} `{}` in `{}`", ud.mode.action_verb(), ud.name, ud.region))
+        }
+        UpgradeState::Failed => {
+            warn!("failed to roll out {}", ud.name);
+            ("danger".into(), format!("failed to {} `{}` in `{}`", ud.mode, ud.name, ud.region))
```
clux: ditto

shipcat_cli/tests/audit.rs
```diff
@@ -0,0 +1,77 @@
...
+        //         "type": "deployment",
+        //         "payload": {
+        //             "id": "svc-v1-egrevision",
+        //             "region": "r1",
+        //             "manifestsRevision": "egrevision",
+        //             "service": "svc",
+        //             "version": "v1"
+        //         }
+        //     }
+        // )))
```
clux: tests not doing much yet?

shipcat_definitions/src/region.rs
```diff
@@ -218,6 +241,54 @@ impl KongConfig {
...
+                if h.token == "IN_VAULT" {
+                    let vkey = format!("{}/shipcat/WEBHOOK_AUDIT_TOKEN", region);
+                    h.token = vault.read(&vkey)?;
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn verify_secrets_exist(&self, vault: &Vault, region: &str) -> Result<()> {
```
clux: will need secrets added as you add this to regions :+1:

shipcat_cli/tests/audit.rs
JoseAlban: not doing enough - but still useful, need to find out a way to freeze timestamp to test payload body properly :/

PR #49, opened 2018-12-04
Allow mutually exclusive features

clux:
This fixes `raftcat` which needs a different deserialize impl atm because it expands it first.


clux:
raftcat deploy fixed by this.

PR #61, opened 2018-12-12
get clusterinfo now requires a cluster parameter

clux:
ambiguities caused in more general clusters


clux:
note the requirement is only there when there is an ambiguity - a region is served by more than one cluster

Issue #11, opened 2018-10-25
remove reliance on tiller


clux:
For a multitude of reasons:

- it makes it harder to inject service mesh functionality (e.g. `conduit inject`)
- gitops style makes the extra source of truth unnecessary and competing

But most importantly; helm has numerous upgrade bugs that makes its use hard to justify:

- https://github.com/helm/helm/issues/3353 : initial install failures cannot be recovered from
- https://github.com/helm/helm/issues/3319 : helm cannot reclaim existing resources (exclusionary)
- https://github.com/helm/helm/issues/4387 : helm sometimes fail to upgrade (forcing us to `helm del --purge`)
- https://github.com/helm/helm/issues/3675 : rollback screws with state (and can cause us to have to purge)
- https://github.com/helm/helm/issues/2735 : helm upgrade doesn't always upgrade

So. Paths to replacing helm means finding ways to replace the things it provides:

- [x] helm hooks (dropped hook use: they are entirely unmanaged resources that are hard to debug, non-idempotent installs) - db migrations should happen in a locked rollout step, or externally
- [x] `helm upgrade --wait --timeout` (shipcat waits instead for now => avoids setting failed state which causes us to have to purge a lot)
- [x] helm history -> kubectl rollout history (slightly worse, but this was never crucial)
- [x] grouping resources for GC purposes -> `kubectl apply --prune -lapp={app} --prune-whitelist=[...]` 
- [x] helm diff plugin -> `kubectl alpha diff` ? (currently in an unusable state - [1.13 GA?](https://github.com/kubernetes/kubernetes/issues/68526))
- [x] helm rollback -> git revert in manifests repo
- [x] shipcat helm upgrade --auto-rollback (useful for rolling environments that gate on integration tests)

NB: `helm template` and chart usage will be kept (for now).

Diff is the only hard issue at the moment that we may have to wait for.

clux:
Diff works fine in kubernetes 1.13: https://github.com/Babylonpartners/shipcat/issues/56

clux:
--auto-rollback flag was removed a few versions back.

The last part of now implementing it is done in #275

clux:
0.127.1 is used internally now with tiller-less mode. Closing.

PR #67, opened 2018-12-18
Allow tiller namespace override via evar

JoseAlban:
Use-case: shipcat was assuming tiller-namespace to be the same namespace in which service would be deployed - and I needed to decouple those concepts in order to use shipcat for deploying on my local kube cluster.

I'm happy to listen why this is not a good idea - but it was immediately useful for me.
Maybe it could be a CLI arg, but tiller is going to disappear soon, so that arg can also be used to invalidate that PR in the future.

Or, this can be wiped out as part of a tiller-removal refactor.


clux:
The idea is to ensure we have a local rbac'd tiller. You can install tiller into the namespace like our cluster standard using: https://github.com/Babylonpartners/shipcat/blob/master/examples/tiller.yml .

JoseAlban:
It's definitely possible to install it like we have on our specific case - but thinking about a more generic approach here, since it's an open-source tool - it'd be nice to be flexible. Unless it's a bad practice to have tiller on `kube-system` for some reason, but for local development it's quite helpful as it minimises custom setup.

clux:
Well, it was kind of bad practice earlier since you couldn't scope RBAC rules. But while I agree it would simplify things a bit, tiller is going away entirely (see https://github.com/helm/community/blob/master/helm-v3/000-helm-v3.md and #11 ) so am inclined not to make changes to tiller integration until that's done.

JoseAlban:
This would soon be obsolete indeed.

PR #69, opened 2019-01-02
Audit webhook: Default coupling to jenkins evars and random UUID as context ID

JoseAlban:
So that we don't need to add a new mapping layer (jenkins evars -> shipcat evars) if we are using jenkins (a sensible default?!).


shipcat_definitions/src/region.rs
```diff
@@ -269,14 +270,22 @@ impl Webhook {
...
+                                .unwrap_or_else(|_| {
+                                    let uuid = Uuid::new_v4().to_string();
+                                    debug!("Assigning random UUID as SHIPCAT_AUDIT_CONTEXT_ID: {}", uuid);
+                                    uuid
+                                }
+                                ));
+
+                whc.insert("SHIPCAT_AUDIT_REVISION".into(),
+                                env::var("SHIPCAT_AUDIT_REVISION")
+                                .or_else(|_| env::var("GIT_COMMIT")) // Jenkins default coupling
```
clux: i'd rather have this in an `if let` like in `slack.rs` to avoid the random coupling deep within an insert statement.

shipcat_definitions/src/region.rs
```diff
@@ -269,14 +270,22 @@ impl Webhook {
...
             Webhook::Audit(_h) => {
-                whc.insert("SHIPCAT_AUDIT_REVISION".into(),
-                                env::var("SHIPCAT_AUDIT_REVISION")
-                                    .map_err(|_| ErrorKind::MissingAuditRevision.to_string())?);
                 whc.insert("SHIPCAT_AUDIT_CONTEXT_ID".into(),
                                 env::var("SHIPCAT_AUDIT_CONTEXT_ID")
-                                    .map_err(|_| ErrorKind::MissingAuditContextId.to_string())?);
+                                .unwrap_or_else(|_| {
+                                    let uuid = Uuid::new_v4().to_string();
+                                    debug!("Assigning random UUID as SHIPCAT_AUDIT_CONTEXT_ID: {}", uuid);
```
clux: remove debug here and just print the struct later on into debug maybe?

shipcat_definitions/src/region.rs
```diff
@@ -289,6 +298,33 @@ impl Webhook {
...
+    use std::env;
+
+    #[test]
+    fn region_webhook_audit_config_jenkins_defaults() {
+        let wha = Webhook::Audit(AuditWebhook{
+            url: Url::parse("http://testnoop").unwrap(),
+            token: "noop".into(),
+        });
+
+        let reuuid = Regex::new(r"\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b").unwrap();
```
clux: a bit over the top test of a dependent library - just test the length with `[0-9a-f-]{36}` or whatever

shipcat_definitions/src/region.rs
```diff
@@ -289,6 +298,33 @@ impl Webhook {
...
+        env::set_var("GIT_COMMIT", "gc1");
+        env::set_var("BUILD_URL", "burl");
+        let cfg = wha.get_configuration().unwrap();
```
clux: maybe we can have this test working without this evar as well? not important though..

shipcat_definitions/src/region.rs
JoseAlban: the intention there was to notify if using an id randomly generated or via environment variable - as this will determine the correlation scope: if generated by shipcat, then narrower scope. But if we print the entire struct later on, we could compare eg vs the jenkins vars and still reach same conclusion indeed - so, will do!

shipcat_definitions/src/region.rs
JoseAlban: yes, can add an extra assertion for it.

shipcat_definitions/src/region.rs
JoseAlban: done!

PR #70, opened 2019-01-03
let slack notifications go to notification channels - fixes #32

### LINKED ISSUE #32: Allow slack notifications to go to notifications channels
Should be an extra call to slack::message or a wrap therein to ensure all calls go to both the configured channel (the one from the evar), and the team's notification channel if it exists.

clux:
technically doubling the notifications atm, because it also goes to the configured evar channel, but good migration start

### DIFF
--- a/tests/shipcat.conf
+++ b/tests/shipcat.conf
@@ -58,7 +58,7 @@ regions:
   webhooks:
     - name: audit
       url: http://testserver/shipcat
-      token: IN_VAULT
+      token: secretsauce
 
 - name: dev-global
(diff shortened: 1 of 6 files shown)


tests/shipcat.conf
```diff
@@ -58,7 +58,7 @@ regions:
   webhooks:
     - name: audit
       url: http://testserver/shipcat
-      token: IN_VAULT
+      token: secretsauce
```
JoseAlban: oh yeah 🍳

Issue #62, opened 2018-12-13
infrastructure syntax for standard use cases


clux:
- [ ] rds
- [ ] s3
- [ ] elasticache

syntax TBD, operator choice dependent.

clux:
abandoning this. service-broker + crossplane are too immature to use atm. keeping infrastructure setups outside manifests.

PR #75, opened 2019-01-10
add basic infrastructure syntax for rds + a reducer - for #62

clux:
basic info. few configurables. reducible to json output via cli.

## usage:
Add to a manifest (say `services/raftcat/shipcap.yml`) by adding:

```diff
+database:
+  engine: postgres
+  version: 9.6
+  size: 20
+  instanceClass: "db.m4.large"
```
## reducing:

```sh
$ shipcat get databases 
---
- name: raftcat
  team: platform
  size: 20
  engine: postgres
  version: "9.6"
  instanceClass: db.m4.large
```
then pass this list to whatever's provisioning this.


shipcat_definitions/src/structs/rds.rs
```diff
@@ -0,0 +1,92 @@
...
+pub enum RdsEngine {
+    Postgres,
+    Mysql
+}
+
+/// RDS Instance types
+///
+/// Subset of the official [AWS RDS instance type list](https://aws.amazon.com/rds/instance-types/).
+#[derive(Deserialize, Serialize, Clone)]
+pub enum InstanceClass {
```
spikelampson: Are we moving to m5 instances any time soon? Need to check on our collection of reserved instances.

clux: added most m5 instances.

clux:
Avoiding this step. Using the abstraction, but because it doesn't neatly plug into anything standard (service broker + crossplane both too immature), it makes no sense to provision these things within kube.

clux:
This is actually still better to have even if the pipeline for it gets split into another repo that generates terraform from it (or what have you). Because the manifests is a better ownership model than two sets of yaml that are linked (but there's no good way of enforcing that).

Issue #71, opened 2019-01-03
crd reconcile needs to create crds if not exists


clux:
Probably best to have this command manage init.

clux:
Was closed by #79

PR #79, opened 2019-01-16
CRD create / update at reconcile

JoseAlban:
Re. #71 .
This ensures that CRDs (shipcatConfig and shipcatManifest) are in the correct state when attempting a `shipcat cluster crd reconcile`.

Tested manually on a local cluster.


shipcat_definitions/src/crds.rs
```diff
@@ -23,6 +24,85 @@ pub struct Metadata {
...
+
+    pub group: String,
+    pub version: String,
+    pub scope: String,
+    pub names: CrdNames,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub additionalPrinterColumns: Option<Vec<CrdAdditionalPrinterColumns>>,
+}
+
+#[derive(Serialize, Deserialize, Clone, Default)]
```
clux: probably shouldn't have default

shipcat_definitions/src/crds.rs
```diff
@@ -23,6 +24,85 @@ pub struct Metadata {
...
+pub struct CrdNames {
+    pub plural: String,
+    pub singular: String,
+    pub kind: String,
+}
+
+#[derive(Serialize, Deserialize, Clone, Default)]
```
clux: also no default

shipcat_definitions/src/crds.rs
JoseAlban: Actually, we need it for `..CrdSpec::default()` ~L70 to infer all other None optionals that might be available.

shipcat_definitions/src/crds.rs
JoseAlban: Same reasoning as above.

PR #81, opened 2019-01-16
[IDM-88] Add support for Kong JWT plugin

janslow:
Converts this:

```yaml
# shipcat.yml
kong:
  auth: jwt # Instead of "oauth2" (default) or "none"
```

into:

```yaml
# output of `shipcat kong`
apis:
  - plugins:
      - name: oauth2
        ensure: removed
      - name: jwt
        ensure: present
        attributes:
          enabled: true
          config:
            uri_param_names: []
            claims_to_verify:
              - exp
            anonymous: ""
```

# Other Kongfig diffs
Changing `kongfig::PluginBase` means that the attributes are no longer required when removing a plugin. Unfortunately, the default value of `ensure` (`present`) is now included for all plugins, as serde doesn't support `skip_serialize_if` on enum variant tags. This should only affect the konfig config, not the end Kong state.

```diff
diff --git a/kong.original.yaml b/kong.yaml
index d2182a0b..bc8136e8 100644
--- a/kong.original.yaml
+++ b/kong.yaml
@@ -5,6 +5,7 @@ apis:
   - name: admin-portal2
     plugins:
       - name: correlation-id
+        ensure: present
         attributes:
           enabled: true
           config:
@@ -21,21 +23,10 @@ apis:
             timeout: 10000
       - name: oauth2
         ensure: removed
-        attributes:
-          enabled: true
-          config:
-            enable_client_credentials: false
-            mandatory_scope: false
-            hide_credentials: false
-            enable_implicit_grant: false
-            global_credentials: true
-            provision_key: XdMdQWOfCUNPh2SkCFNPBAxj4liJ81Lf5KDcvpRvYA
-            enable_password_grant: true
-            enable_authorization_code: true
-            anonymous: ""
-            token_expiration: 1800
-            accept_http_if_already_terminated: false
```

All APIs now have the JWT plugin removed:

```diff
@@ -283,6 +263,8 @@ apis:
             anonymous: ""
             token_expiration: 1800
             accept_http_if_already_terminated: false
+      - name: jwt
+        ensure: removed
     attributes:
       hosts: {}
       uris:
```


shipcat_cli/tests/common.rs
```diff
@@ -246,31 +266,22 @@ fn kong_test() {
...
+    assert_eq!(attr.enabled, true);
+    assert_eq!(attr.config.header_name, "babylon-request-id");
 
-    if let kongfig::ApiPlugin::TcpLog(p) = &api.plugins[1] {
-        assert!(p.ensure == kongfig::Ensure::Present);
-        assert_eq!(p.attributes.enabled, true);
-    } else {
-        panic!("plugin 1 is not a tcp-log plugin")
-    }
+    let attr = plugin_attributes!("TcpLog", &api.plugins[1], kongfig::ApiPlugin::TcpLog);
```
clux: i kind of think the previous is more readable here, but it's fine with me - but fine if it's in its own file.

shipcat_definitions/src/structs/kong.rs
```diff
@@ -121,6 +121,38 @@ impl Kong {
...
+            }
+            Authentication::None => {
+                if let Some(_) = self.oauth2_anonymous {
+                    bail!("`oauth2_anonymous` not supported when Kong `auth` is `none`");
+                }
+                if let Some(true) = self.oauth2_extension_plugin {
+                    bail!("`oauth2_extension_plugin` not supported when Kong `auth` is `none`");
+                }
+            }
+        }
```
clux: ah, these are good :+1:

shipcat_definitions/src/structs/kongfig.rs
```diff
@@ -174,199 +146,176 @@ pub struct Oauth2PluginAttributesConfig {
...
-pub struct PluginBase<T> {
-    #[serde(skip_serializing_if = "Ensure::is_present")]
-    pub ensure: Ensure,
-    pub attributes: T,
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(tag = "ensure", content = "attributes", rename_all = "lowercase")]
+pub enum PluginBase<T> {
+    Present(PluginAttributes<T>),
+    Removed,
 }
```
clux: oh this is how you've done the magic, that's neat.

clux:
Would you like to have a better way to rollout shipcat versions across environments for stuff like this?

shipcat_cli/tests/common.rs
```diff
@@ -19,6 +19,26 @@ macro_rules! btree_set {
...
+
+macro_rules! assert_plugin_removed {
+    ( $name:expr, $plugin:expr, $type:path ) => {
+        match $plugin {
+            $type(kongfig::PluginBase::Removed) => {},
+            $type(_) => panic!("{} plugin is not removed", $name),
+            _ => panic!("plugin is not a {} plugin", $name),
+        }
+    };
+}
```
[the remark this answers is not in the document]
janslow: I've split it out.

janslow:
Yeah, that would be handy.

Alternatively, decoupling the Kong output from Shipcat, so that you can specify the "Kong config version" on a per region/environment basis.

This would also be useful for the Helm charts, so that you can make a change to the charts in `dev` first, then roll it out to other environments in order.

shipcat_cli/tests/common.rs
janslow: I generally aim to keep the complexity of the test assertions to a minimum

clux:
> Would you like to have a better way to rollout shipcat versions across environments for stuff like this?

clux:
oops

janslow:
Ok, ready to merge.

clux:
Created https://github.com/Babylonpartners/shipcat/issues/85 to start addressing better rollout strategy.

PR #83, opened 2019-01-17
Allow ShipcatConfig CRD to be deployed without filtering on a cluster

JoseAlban:
This enables a cluster to be configured eg on `shipcat.conf`:
```
- name: dev-ops
  namespace: dev
  environment: dev
[...]
  customResource:
    config: RegionAgnosticBase
```

Which causes a shipcatConfig instance named `region-agnostic` to be deployed on the cluster, and contains information about all other available clusters/regions (ie it's unfiltered).

It's a replacement (ie on this eg, `region-agnostic` is deployed iof `dev-ops` instance), rather than addition.


shipcat_cli/src/cluster.rs
```diff
@@ -47,7 +47,13 @@ fn crd_reconcile(svcs: Vec<String>, config: &Config, region: &Region, n_workers:
     use std::sync::mpsc::channel;
 
     // Make sure config can apply first
-    kube::apply_crd(&region.name, config.clone(), &region.namespace)?;
+    let applycfg = if let Some(ref crs) = &region.customResource {
+        // special configtype detected - re-populating config object
+        Config::new(crs.shipcatConfig.clone(), &region.name).unwrap().0
```
clux: ? rather than unwrap

clux: might want to pass in the correct config before - but not sure how much harder that is
this is a bit non-standard atm

shipcat_cli/tests/common.rs
```diff
@@ -47,12 +47,22 @@ fn config_test() {
...
+    let (conf, _region) = filteredcfg.unwrap(); // better to unwrap and get full trace
     assert!(Config::new(ConfigType::File, "dev-uk").is_err());
     assert!(conf.print().is_ok());
 }
 
+#[test]
+fn config_cr_settings_test() {
+    setup();
+    Config::read().unwrap(); // iof assert!(Config::read().is_ok());
+    assert!(Config::new(ConfigType::Base, "dev-ops").is_ok());
```
clux: just unwrap here, better trace

shipcat_definitions/src/config.rs
```diff
@@ -351,14 +351,14 @@ impl Config {
...
-        } else {
-            bail!("Config::new only supports Completed or Base type");
+        } else if kind != ConfigType::RegionAgnosticBase {
+            bail!("Config::new only supports Filtered, Base and RegionAgnosticBase types");
         }
 
-        if kind == ConfigType::Completed {
-            conf.complete(&region)?;
+        if kind == ConfigType::Filtered {
+            conf.fill_secrets(&region)?;
```
clux: thing manifest equivalent here is just called `secrets`

shipcat_definitions/src/region.rs
```diff
@@ -405,6 +413,8 @@ pub struct Region {
     pub locations: Vec<String>,
     /// All webhooks
     pub webhooks: Option<Vec<Webhook>>,
+    /// CRD tuning
+    pub customResource: Option<CRSettings>
```
clux: `customResources`, probably
manifests doesn't need a flag yet, but dunno if we should have a marker for it?

shipcat_definitions/src/states.rs
```diff
@@ -114,14 +114,16 @@ impl Manifest {
...
-    /// This is the upgraded version of a Filtered config for the same region.
-    /// Secrets have been resolved here.
-    Completed,
+    /// A filtered config for a specific region, with resolved secrets
+    Filtered,
+
+    /// Region-independent, unresolved secrets
+    /// 
+    /// Just like Base - but for all regions
+    RegionAgnosticBase,
```
clux: `UnionisedBase`

JoseAlban: good point - for future's sake

shipcat_definitions/src/region.rs
JoseAlban: nw will pluralise - and add a default boolean-true for `manifests`

shipcat_cli/src/cluster.rs
JoseAlban: Yeah - ideally we'd have a `Config.new` only returning the actual config, and a potential `Region.new` to get the CRD configuration before `Config.new`ing it. But would defer that refactor for the sake of minimum amount of LOC for feature on this PR. We can bring this up later on a diff PR if needed.

shipcat_definitions/src/crds.rs
```diff
@@ -125,9 +125,13 @@ impl From<Config> for Crd<Config> {
...
         let rgs = conf.list_regions();
         assert!(!conf.has_secrets()); // no secrets
-        assert_eq!(rgs.len(), 1); // config must be filtered
-        // thus, can infer the region :-)
-        let rname = rgs[0].to_owned();
+        let rname: String = if rgs.len() == 1 { // config has been filtered
+            // thus, can infer the region :-)
+            rgs[0].to_owned()
+        } else { // non-filtered
+            "unionised".to_owned()
```
clux: may want to make sure that a region name isn't this as well :(

clux: i.e. a thing in `Region::verify`

JoseAlban: :| indeed.. good point - will validate

PR #89, opened 2019-01-23
fix demo chart after env changes (minikube)

clux:


nitesh-babylon:
I ran this locally and it no longer is throwing an error

PR #92, opened 2019-01-25
elasticache provisioning + reducers

clux:
basic info. few configurables. reducible to json output via cli.

## Usage
Add to a manifest (say services/raftcat/shipcap.yml) by adding:

```diff
+redis:
+  nodes: 2
+  nodeType: cache.m4.large
```

## Reducing

```yaml
$ shipcat get caches
---
- name: raftcat
  team: platform
  nodes: 2
  nodeType: cache.m4.large
```

then pass this list to whatever's provisioning this.

## Extra
Also removes a hidden implicit for `instanceClass` (rds module) because the same pattern in here made it obvious that it should be explicit.

### DIFF
--- a/shipcat_definitions/src/structs/elasticache.rs
+++ b/shipcat_definitions/src/structs/elasticache.rs
@@ -8,25 +8,36 @@ use super::Metadata;
 /// ElastiCache Node Types
 ///
 /// Subset of the official [AWS ElastiCache node type list](https://aws.amazon.com/elasticache/pricing/).
+/// Only current generation (m5 + r5) + along with cheap t2 nodes
 #[derive(Deserialize, Serialize, Clone)]
 pub enum NodeType {
+    // Standard t2 nodes
     #[serde(rename = "cache.t2.micro")]
-    CacheT2Micro,
-
+    CacheT2micro,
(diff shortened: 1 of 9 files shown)


shipcat_definitions/src/structs/elasticache.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// Sometimes referred to as num cache clusters (in cluster mode disabled).
+    /// This number includes the master/shard.
+    ///
+    /// Must be between an integer `1` and `6`.
+    pub nodes: Option<u8>,
+
+    /// The node type of the ElastiCache instance
+    ///
+    /// Defaults to cache.m4.large
+    pub nodeType: Option<NodeType>,
```
krubot: Should we be defaulting to a smaller type? Two node cluster at `cache.m4.large` costs on average $132 while a `cache.t2.medium` costs on average $56.

spikelampson: Do we have any reserved instances? If so, we should defo default to them...

clux: I've removed the default. The documentation was leftover.

Supporting more choice now.

PR #93, opened 2019-01-26
Add global `--region` option

### LINKED ISSUE #91: CLI option to set Kubectl context
Currently Shipcat executes `kubectl config current-context` to discover the current context, which it uses to choose the region.

It would be useful if there was a way to override this, e.g., a `--context` option (which matches the `kubectl` option) or `--region`.

janslow:
Instead of declaring `--region` on each subcommand (as it's currently missing from quite a few), make it a global option. Fixes #91.

Also, make `--debug` and `--verbose` options global. This allows them to be used anywhere in the command. Previously, the following wasn't allowed:

```bash
$ shipcat kong --debug
error: Found argument '--debug' which wasn't expected, or isn't valid in this context
```

Finally, `Cargo.lock` wasn't committed last time the version was bumped, so fix that.


shipcat_cli/src/main.rs
```diff
@@ -155,11 +153,6 @@ fn main() {
               .about("Reduce data handling structs"))
 
         .subcommand(SubCommand::with_name("get")
-              .arg(Arg::with_name("region")
-                .short("r")
-                .long("region")
-                .takes_value(true)
-                .help("Specific region to check"))
```
clux: removing this from commands is going to break a bunch of existing references and jobsr. not sure if we can do this immediately.

maybe a two step procedure. i need to check tomorrow.

shipcat_cli/src/main.rs
```diff
@@ -32,11 +32,19 @@ fn main() {
...
             .short("d")
             .long("debug")
+            .global(true)
             .help("Adds line numbers to log statements"))
+        .arg(Arg::with_name("region")
+                .short("r")
+                .long("region")
+                .takes_value(true)
+                .global(true)
+                .help("Region to use (dev-uk, staging-uk, prod-uk)"))
```
clux: oh does global actually make it work in any position?

it certainly makes sense as a global arg, even though there are a few subcommands it conflicts with (but that's fine).

janslow: Yes, it propagates the arguments down to all subcommand when configuring (but propagates the value back up as well).

As a result, `shipcat --region=dev-uk kong` is now the same as `shipcat kong --region=dev-uk`, if you use [`global(true)`](https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.global).

PR #98, opened 2019-02-05
Add support for k8s jobs

jackatbancast:
Adds support for k8s jobs a la https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/


shipcat_definitions/src/structs/job.rs
```diff
@@ -0,0 +1,102 @@
...
+///
+/// Used to decide if a job should be restarted when it fails or not.
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub enum RestartPolicy {
+    Never,
+    OnFailure,
+}
+
+impl Default for RestartPolicy {
+    fn default() -> Self { RestartPolicy::Never }
```
clux: is this the kube default?

examples/charts/base/templates/jobs.yaml
```diff
@@ -0,0 +1,63 @@
...
+    type: {{ $.Values.type | default "job" }}
+{{- if $.Values.labels }}
+{{ toYaml $.Values.labels | indent 4 }}
+{{- end }}
+    chart: {{ template "chart.chart" $ }}
+    release: {{ $.Release.Name }}
+    heritage: {{ $.Release.Service }}
+spec:
+  concurrencyPolicy: Forbid
+  schedule: "{{ $v.schedule }}"
```
clux: how about a backoff?

shipcat_definitions/src/manifest.rs
```diff
@@ -594,6 +593,18 @@ pub struct Manifest {
...
     pub cronJobs: Vec<CronJob>,
 
+    /// Job images to run as kubernetes `Job` objects
+    ///
+    /// An abstraction on top of [kubernetes jobs](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/)
+    ///
+    /// ```yaml
+    /// jobs:
+    /// - name: run-migration
+    ///   command: ["bundle", "exec", "rake", "db:migrate"]
```
clux: hm, are you advocating for db migration in job objects along with a service?
jobs would run along with rolling upgrades, right?

shipcat_definitions/src/manifest.rs
```diff
@@ -594,6 +593,18 @@ pub struct Manifest {
...
+    /// ```
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub jobs: Vec<Job>,
```
clux: There are some assumptions in shipcat around the idea that a manifest abstracts a deployment object. Like `httpPort` => service etc. `ports` are for a deployment+service

Are there any other awkward bits around this for jobs? Like, is this meant to be able to run by themselves? What would a minimal job-only manifest contain? What's the lifecycle management of something like that in manifests?

shipcat_definitions/src/structs/job.rs
jackatbancast: It must be specified.
> In addition to required fields for a Pod, a pod template in a job must specify appropriate labels (see pod selector) and an appropriate restart policy.

shipcat_definitions/src/manifest.rs
jackatbancast: Jobs would run alongside. This would work with migrations that are either specified by version (migrations/code should be backwards compatible anyway). It would also work with something that uses an external store (e.g. S3/GitHub) as a source for migrations.

clux: i'd still have some non-migration code as an example here, it's not great to recommend this approach tbh.

shipcat_definitions/src/manifest.rs
clux: it appears you're just re-using properties for the deployment for most things. maybe worthy of a comment of what's reused?

shipcat_definitions/src/structs/job.rs
```diff
@@ -0,0 +1,102 @@
...
+    pub image: Option<String>,
+
+    /// Version to use for overridden images
+    ///
+    /// Only allowed if image is set.
+    pub version: Option<String>,
+
+    /// Resource limits and requests
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub resources: Option<Resources<String>>,
```
clux: resource calculation ignored for this

clux: although don't know how you would calculate this sensibly..

shipcat_definitions/src/structs/job.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+
+impl Job {
+    pub fn verify(&self) -> Result<()> {
+        let re = Regex::new(r"^[0-9a-z\-]{1,50}$").unwrap();
+        if !re.is_match(&self.name) {
+            bail!("Please use a short, lower case job names with dashes");
+        }
+        // TODO: version verify
```
clux: maybe implement this

shipcat_definitions/src/structs/job.rs
```diff
@@ -0,0 +1,102 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Default)]
+#[serde(deny_unknown_fields)]
+pub struct JobVolumeClaim {
+    /// The cron job name
+    pub size: String,
+    pub mountPath: String,
+}
+
+#[derive(Serialize, Deserialize, Clone, Default)]
```
clux: no default here

shipcat_definitions/src/structs/job.rs
```diff
@@ -0,0 +1,102 @@
...
+pub enum RestartPolicy {
+    Never,
+    OnFailure,
+}
+
+impl Default for RestartPolicy {
+    fn default() -> Self { RestartPolicy::Never }
+}
+
+#[derive(Serialize, Deserialize, Clone, Default)]
```
clux: no default

shipcat_definitions/src/structs/job.rs
```diff
@@ -84,12 +79,15 @@ impl Job {
         if !re.is_match(&self.name) {
             bail!("Please use a short, lower case job names with dashes");
         }
-        // TODO: version verify
+        if let Some(v) = &self.version {
+            let semver_re = Regex::new(r"v?\d+\.\d+\.\d+").unwrap();
+            if !semver_re.is_match(&v) {
+                bail!("Please use a valid semver tag for the image version");
+            }
```
clux: please use the semver crate we use everywhere!

jackatbancast: Done ✅

PR #100, opened 2019-02-05
Manage vault policies for teams

clux:
## Intro
Vault secrets live inside flat region folders (based on Region in shipcat.conf) where the folders are partitioned by owning teams in `shipcat.conf`.

We would like to have a way to automatically make team admins able to have "create" and "list" access to any vault (even production) so that they can at least insert things themselves.

## Proposed abstraction
Generate a policy from shipcat, customizable via the `Team` struct in shipcat.conf.
Can be used to generate the policy directly (fake data):

```sh
$ shipcat get vault-policy AIM
shipcat::get: Team 'AIM' does not define a vaultAdmins team in shipcat.conf
path "sys/*" {
  policy = "deny"
}
path "secret/*" {
  capabilities = ["list"]
}
path "secret/*/aim-dashboard/*" {
  capabilities = ["create", "list"]
}
path "secret/*/aim-models/*" {
  capabilities = ["create", "list"]
}
```

and then we can associate that list policy with a github admin team if it is set in `shipcat.conf`.

## Automagic
Automatically generate policies for all teams that define an admin team, and associate the policies with the team in vault auth:

```sh
$ shipcat cluster vault-policy reconcile
shipcat::cluster: Starting N parallel vault jobs using 8 workers
shipcat::cluster: Team 'AIM' does not have a defined vaultAdmins team in shipcat.conf - ignoring
shipcat::cluster: Writing vault policy for developer-experience to ./developer-experience-policy.hcl
shipcat::cluster: Applying vault policy for developer-experience
Success! Uploaded policy: developer-experience
shipcat::cluster: Associating vault policy for DevExperience with github team: developer-experience
Success! Data written to: auth/github/map/teams/developer-experience
```

Can run daily anywhere that's secure enough to house policy writing credentials.

## Caveats + Unresolveds

- Probably need better GC of moved github teams (currently the old policies would persist)
- Team validation?
- Little customisation of capabilities: `["create", "list"]` only for now - immutable + does not escalate to raw db access
- Needs a hefty acc. to run this reconcile
~- Maybe rename to `shipcat cluster vault-policy reconcile` ?~
- Store the policies in git as a human readable step? (at least mark them as autogenerated then)
- Use templates for this? They're currently really small scripts.
- Is there an HCL validator we can use on the output?
- The ANY REGION folder is currently using a wildcard in the policy to make it region agnostic. This was to avoid a chicken-egg problem (see code)


JoseAlban:
>    * Maybe rename to `shipcat cluster vault-policy reconcile`

yes please - otherwise it might comm that we're actually reconciling secrets as well, and if so, it better be a diff cmd anyway

shipcat_cli/src/get.rs
```diff
@@ -71,6 +71,58 @@ pub fn codeowners(conf: &Config) -> Result<Vec<String>> {
...
+    //    // default policy only
+    //    return Ok(output)
+    //};
+
+    for svc in Manifest::all()? {
+        // Can rely on blank here because metadata is a global property
+        let mf = Manifest::blank(&svc)?;
+        if let Some(md) = mf.metadata {
+            if md.team == team_name {
+                // allow this team to manage this folder <- later
```
JoseAlban: change `<- later` to `TODO` ?

shipcat_cli/src/get.rs
```diff
@@ -71,6 +71,58 @@ pub fn codeowners(conf: &Config) -> Result<Vec<String>> {
...
+                let base_perms = "[\"create\", \"list\"]";
+                // NB: extra star is the mandatory vault.folder (varies per environment)
+                output += &format!("path \"secret/*/{}/*\" {{\n  capabilities = {}\n}}\n", mf.name, base_perms);
+            }
+        }
+    }
+    println!("{}", output);
+    Ok(output)
+}
+
```
JoseAlban: Add at least a known-to-work vault policy payload as a test here - so safeguard against misplaced `\n`s etc

shipcat_definitions/src/config.rs
```diff
@@ -58,10 +58,19 @@ pub struct Cluster {
...
+    ///
+    /// Used to generate CODEOWNERS merge policies.
     #[serde(default)]
     pub owners: Vec<Contact>,
+
+    /// Admin team on github
+    ///
+    /// Used to generate vault policies
     #[serde(default)]
+    pub admins: Option<String>,
```
JoseAlban: call it `adminTeams` rather? so that it's clear it's not for individual gh handlers

clux:
have added basic tests, and accounted for comments. will add some team name validation by a regex then probably merge if it looks ok.

clux:
All done.

PR #116, opened 2019-02-27
raftcat upgradeability + small fixes

clux:
- adds prometheus client to raftcat
- makes `deny_unknown_fields` a `shipcat_cli` feature only (i.e. not `raftcat`)
- bugfix on external resource breaking codeowners generation


shipcat_definitions/src/region.rs
```diff
@@ -133,7 +133,7 @@ impl VaultConfig {
 
 /// Kafka configuration for a region
 #[derive(Serialize, Deserialize, Clone, Default)]
-#[serde(deny_unknown_fields)]
+#[cfg_attr(filesystem, cfg_attr(filesystem, serde(deny_unknown_fields)))]
```
JoseAlban: Apart from these typo'es, all good

PR #117, opened 2019-03-07
vault policies now generated from templates - fixes #107

### LINKED ISSUE #107: vault policy templates
To make it easier to customise vault policies, they should just be read from:

- `vault/team.policy.hcl.j2`
- `vault/team.prod.policy.hcl.j2`

And templated by `shipcat get vault-policy`.

clux:


shipcat_definitions/src/template.rs
```diff
@@ -152,3 +152,44 @@ impl EnvVars {
         Ok(())
     }
 }
+
+
+/// Read an arbitrarty template from manifests/{folder}/{name}.j2
```
gleaver: `arbitrarty`

PR #125, opened 2019-03-14
raftcat fixes

clux:
long overdue


raftcat/src/main.rs
```diff
@@ -346,11 +356,19 @@ fn main() -> Result<()> {
     let dsn = env::var("SENTRY_DSN").expect("Sentry DSN required");
     let _guard = sentry::init(dsn); // must keep _guard in scope
 
-    env::set_var("RUST_LOG", "actix_web=info,raftcat=info,kubernetes=info");
-    //env::set_var("RUST_LOG", "actix_web=info,raftcat=debug");
-    //env::set_var("RUST_BACKTRACE", "full");
+    if let Ok(level) = env::var("LOG_LEVEL") {
+        if level.to_lowercase() == "debug" {
```
janslow: Won't `RUST_LOG` be unset if `LOG_LEVEL=info` or similar?

clux: good spot, thanks.

Issue #120, opened 2019-03-11
Split manifest parsing from main Manifest struct


janslow:
Currently, the `Manifest` struct (and all of the structs which it references, e.g., `Kong`) is used for parsing/merging the files for a service (e.g., `shipcat.yml`, `dev-uk.yml`) and for representing an assembled/validated manifest for a service (e.g., to write to a CRD).

For the assembled manifest, this means:
* There are backwards-compatibility properties which need to be handled (e.g., the `Kong` has `hosts`, `host` and `uris` properties)
* Defaulted properties are still `Option`s (e.g., `Kong#auth`, so callers need to manually unwrap it to the default value).

For the sources, this means:
* All source files are allowed to be the same shape (e.g., `dev-uk.yml` can define `name`, so that has to be manually validated)
* Adding alternative syntaxes for common features (e.g., allowing `env: { MY_VAR: ~ }` to delete an env var) adds complexity to the Manifest type, which affects the entire codebase.

# Proposal

* `shipcat.yml` and the overrides (e.g., `dev-uk.yml`) are passed as `ServiceDefaults`/`ServiceOverrides`.
  * Simple validation is handled by parser (e.g., there is no `name` field in `ServiceOverrides`, so it's automatically invalid in `dev-uk.yml`)
* The region/global defaults from `shipcat.conf` are parsed as `GlobalDefaults`, `RegionDefaults`.
* The above sources are assembled into a `Manifest`
  * Required properties can be safely obtained from default values `region.replicaCount.or(env.replicaCount).or(service.replicaCount).expect("replicaCount is required") (we'd probably want a nicer syntax).
  * Additional validation (e.g., env vars are upper case) are also performed
  * More complex conversion occurs (e.g., parsing env var templates)
* `Manifest` structs are passed to the rest of Shipcat
  * They are always valid (e.g., required properties are not `Option`s)
  * Only high level validation (e.g., is a secret missing) is performed by commands.

_I'm not convinced with the above naming conventions, open to suggestions._

clux:
Yeah, that will take away a lot of the magic 50+ `option.unwrap()` calls that is guaranteed by verify, which would definitely make the codebase nicer. I strayed away from it initially because it felt like everything in what is `Manifest` had to be duplicated on what you call `ServiceOverrides` as optionals anyway.

The kong host/hosts issue is a bit messy. Definitely some of those fields should live in a `gate:` syntax that Themis started, as they don't really related to kong at all.

[...]

PR #126, opened 2019-03-18
[IDM-104] Extract reading manifests from files from `shipcat_definitions`

janslow:
Reading and merging manifests from disk has now been moved into `shipcat_filebacked`.

This is the first step towards #120. Currently, I've only replaced `Manifest` with:

* `ManifestSource` (used to parse `shipcat.yml`)
* `ManifestOverrides` (used to parse `dev-uk.yml`/`prod.yml` etc)
* `ManifestDefaults` (built from regional/globla defaults in `shipcat.conf`).

Most of these changes are just pushing code down from `shipcat_definitions` to `shipcat_filebacked` (and changing `shipcat` to use the new methods). Cleaning up the `Manifest` struct (e.g., `image` and `imageSize` shouldn't be optional) can be done in a future PR.


merge_derive/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
```
janslow: Procedural macros have to be defined in a dedicated package.

tests/services/fake-ask/shipcat.yml
```diff
@@ -1,9 +1,9 @@
 name: fake-ask
 labels:
-  custom-metrics: true
+  custom-metrics: 'true'
```
janslow: I've no idea why this worked originally (it used to coerce non-strings into strings).

janslow:
Before merging, I want to run this against a few real environments and do a diff of the Helm output and CRDs.

merge/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "merge"
+version = "0.92.0"
+authors = ["Jay Anslow <jay@anslow.me.uk>"]
+edition = "2018"
```
clux: sweet. publishing this to crates?

merge_derive/Cargo.toml
clux: omg you're doing derived merge! i was hoping we'd eventually get to that point but it seemed so complicated. awesome that you're trying this :heart:

shipcat_filebacked/src/manifest.rs
```diff
@@ -0,0 +1,327 @@
...
+#[serde(default, rename_all = "camelCase")]
+pub struct ManifestSource {
+    pub name: Option<String>,
+    pub external: Option<bool>,
+    pub disabled: Option<bool>,
+    pub regions: Option<Vec<String>>,
+    pub metadata: Option<Metadata>,
+
+    #[serde(flatten)]
+    pub overrides: ManifestOverrides,
```
clux: woah, flatten overrides. this is awesome!

shipcat_filebacked/src/manifest.rs
```diff
@@ -0,0 +1,327 @@
...
+            // TODO: Must be non-empty
+            regions: self.regions.unwrap_or_default(),
+            metadata: metadata,
+            chart: defaults.chart,
+            // TODO: Make imageSize non-optional
+            imageSize: overrides.image_size.or(Some(512)),
+            // TODO: Make image non-optional
+            image: Some(image),
+            version: overrides.version,
+            command: overrides.command.unwrap_or_default(),
```
clux: guess people are forced to make new lines in here when making syntax, which is much better than to "be informed to copy lines into merge.rs".

tests/services/fake-ask/shipcat.yml
clux: hm, that's strange, it should coerce.

guess otherwise you have a less fun manifest pr ahead of you :(

tests/services/fake-ask/shipcat.yml
```diff
@@ -1,9 +1,9 @@
...
 resources:
   limits:
-    cpu: 2
+    cpu: '2'
```
clux: oh this as well :<

shipcat_filebacked/src/manifest.rs
```diff
@@ -0,0 +1,327 @@
...
+    pub labels: BTreeMap<String, String>,
+    pub kong: Option<Kong>,
+    pub gate: Option<Gate>,
+    pub hosts: Option<Vec<String>>,
+    pub kafka: Option<Kafka>,
+    pub source_ranges: Option<Vec<String>>,
+    pub rbac: Option<Vec<Rbac>>,
+
+    #[serde(flatten)]
+    pub defaults: ManifestDefaults,
```
clux: how does flatten interact with existing keys?

janslow: It's just treated the same as any other field (e.g., `rbac`), the derive merge of `ManifestOverrides` calls `ManifestDefaults#merge` (which is provided by the derive on `ManifestDefaults`).

```rust
impl Merge for ManifestDefaults {
  pub fn merge(self, other: ManifestDefaults) -> ManifestDefaults {
    ManifestDefaults {
		..., // Other properties
		defaults: self.defaults.merge(other.defaults),
		// Equivalent to
		defaults: ManifestDefaults {
			chart: self.defaults.chart.merge(other.defaults.chart),
			...,
		}
    }
}
```

shipcat_filebacked/src/manifest.rs
janslow: Yep, more safety at compile time.

tests/services/fake-ask/shipcat.yml
janslow: It's triggered by putting `resources` in `ManifestOverrides` which is then flattened.

janslow: Now fixed by introducing a `RelaxedString` which can read strings, numbers, booleans and nulls.

janslow:
I've compared the outputs (`shipcat helm $service template` (dev only), `shipcat crd $service` and `shipcat kong` for each service) of Shipcat `0.92.0` and the build of this one, there are no differences.

shipcat_definitions/src/relaxed_string.rs
```diff
@@ -0,0 +1,114 @@
+use std::fmt;
```
janslow: After more logic/structs is pulled into `shipcat_filebacked`, it should be possible to pull this module down into that package.

merge/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Values defined in `other` take precedence over values defined in `self`.
+    fn merge(self, other: Self) -> Self;
+}
+
+impl<T> Merge for Option<T> {
+    #[inline]
+    fn merge(self, other: Self) -> Self {
+        other.or(self)
+    }
+}
```
clux: i guess you're not doing an impl for Vector here because it's all wrapped in options now?

shipcat_cli/src/get.rs
```diff
@@ -182,7 +182,7 @@ pub fn apistatus(conf: &Config, reg: &Region) -> Result<()> {
 
     // Get API Info from Manifests
     for svc in Manifest::available(&reg.name)? {
-        let mf = Manifest::simple(&svc, &conf, &reg)?;
+        let mf = shipcat_filebacked::load_manifest(&svc, &conf, &reg)?;
```
clux: does this now read templates for manifests in this case? because i found that slowed down gate generation stuff quite a lot. the simple distinction was useful for some of the crazier `get` functions because they were used a lot in validation scripts (that iterate over every region)

shipcat_definitions/src/structs/cronjob.rs
```diff
@@ -34,7 +35,7 @@ pub struct CronJob {
 
     /// Resource limits and requests
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub resources: Option<Resources<String>>,
+    pub resources: Option<Resources<RelaxedString>>,
```
clux: does math stuff still work? i thought that had conversions from String -> f64

shipcat_definitions/src/structs/resources.rs
```diff
@@ -43,16 +44,16 @@ pub struct Resources<T> {
...
-            memory: parse_memory(&self.requests.memory)?,
-            cpu: parse_cpu(&self.requests.cpu)?,
+            memory: parse_memory(&self.requests.memory.to_string())?,
+            cpu: parse_cpu(&self.requests.cpu.to_string())?,
         };
         let limits = ResourceLimit {
-            memory: parse_memory(&self.limits.memory)?,
-            cpu: parse_cpu(&self.limits.cpu)?,
+            memory: parse_memory(&self.limits.memory.to_string())?,
+            cpu: parse_cpu(&self.limits.cpu.to_string())?,
```
clux: ah, it's a simple changeover

shipcat_filebacked/src/load.rs
```diff
@@ -0,0 +1,117 @@
...
+            manifest = manifest.merge_overrides(env);
+        }
+
+        let region_path = dir.join(format!("{}.yml", reg.name));
+        if region_path.is_file() {
+            let region = ManifestOverrides::read_from(&region_path)?;
+            manifest = manifest.merge_overrides(region);
+        }
+
+        manifest.build(&conf, &reg)
```
clux: so looks like this is equivalent to Simple ? don't see any template / secret handling in here

shipcat_filebacked/src/manifest.rs
```diff
@@ -0,0 +1,332 @@
...
+    // TODO: Extract ConfigsSource
+    fn build_configs(&self, service: &String) -> Result<Option<ConfigMap>> {
+        let original = &self.overrides.configs;
+        if original.is_none() {
+            return Ok(None);
+        }
+        let mut configs = original.clone().unwrap();
+        for f in &mut configs.files {
+            f.value = Some(read_template_file(service, &f.name)?);
+        }
```
clux: ah template handling is a build operation now :+1:

clux:
thankfully, compile times on the cli don't seem to have taken much of a hit, but raftcat does not compile:

```
error[E0433]: failed to resolve: use of undeclared type or module `relaxed_string`
  --> shipcat_definitions/src/structs/env.rs:39:85
   |
39 |     #[serde(default, skip_serializing_if = "BTreeMap::is_empty", deserialize_with = "relaxed_string::deserialize")]
   |   
```

clux:
it is breaking validation in manifests, do you have a branch there?
```
shipcat: validate error: invalid type: boolean `false`, expected a string at line 1 column 5
```

shipcat_cli/src/get.rs
janslow: Ok, I'll create an equivalent version of `simple`

merge/src/lib.rs
janslow: Yep.

So the only things that are merged are `Option`, `BTreeMap` and structs which derive merge.

tests/services/fake-ask/shipcat.yml
janslow: Now fixed

shipcat_cli/src/get.rs
janslow: Introduced `SimpleManifest` which is a subset of `Manifest`, removing the need for the `ManifestType::Simple`.

clux:
Have done some benchmarking, compile times just affected slightly on full build - but indistinguishable when changing individual files. `get` calls seems fast still, raftcat works (have better ideas of where to add tests now), validate works in pr.

provided a diff looks sane gogogo.

new crates are a nice refactor.

PR #129, opened 2019-03-19
Add properties to kafka struct

lostick:


shipcat_definitions/src/region.rs
```diff
@@ -119,6 +119,9 @@ pub struct KafkaConfig {
     ///
     /// These are injected in to the manifest.kafka struct if it's set.
     pub brokers: Vec<String>,
+
+    /// A mapping of kafka properties to environment variables (optional)
+    pub propertyEnvMapping: BTreeMap<String, String>,
```
clux: stick a `#[serde(default)]` on this please

lostick: well spotted

clux: actually can this be put under a kafka struct in `Region`?

lostick:
@clux I've bumped `shipcat` to `0.93.0`

PR #134, opened 2019-03-26
raftcat code improvements

clux:
- raftcat version in templates
- state management now handled in own file
- state file encapsulates locking behaviour (so main.rs doesn't have to)


clux:
other than the extra template addition, there's no functional changes

clux:
shoehorned in one addition with `shipcat apply --no-wait` to mirror the old `shipcat helm upgrade --no-wait`

PR #137, opened 2019-03-27
shipcat diff

### LINKED ISSUE #56: Add a smarter difference system
[derive-diff](https://github.com/polachok/derive-diff/blob/master/test-derive-diff/src/lib.rs) or [difference](https://github.com/johannhof/difference.rs) crates can let us spot the exact changes we are making either when:

- we get an updated crd in raftcat
- `shipcat diff service` which can diff against the CRD

This allows us to have a working diff withouth needing kube 1.13. Plus it can be used to check what you change for pre-merge checks in manifests.

clux:
fixes #56


clux:
Ok, both modes work now.

```
$ shipcat diff admiral-podrington --git
shipcat::kube: kubectl config current-context
shipcat_filebacked::load: Loading service manifest from "./services/admiral-podrington/shipcat.yml"
shipcat_filebacked::load: Loading service overrides from "./services/admiral-podrington/dev-ops.yml"
shipcat::diff: kubectl checkout master --quiet
shipcat::diff: kubectl stash --quiet
shipcat_filebacked::load: Loading service manifest from "./services/admiral-podrington/shipcat.yml"
shipcat_filebacked::load: Loading service overrides from "./services/admiral-podrington/dev-ops.yml"
shipcat::diff: kubectl stash pop --quiet
shipcat::diff: kubectl checkout - --quiet
```
then prints a github style diff.

the other one (default) relies on kubernetes 1.13 and kubectl diffs the crd:

```sh
$ shipcat diff admiral-podrington -v
shipcat::kube: kubectl config current-context
shipcat_filebacked::load: Loading service manifest from "./services/admiral-podrington/shipcat.yml"
shipcat_filebacked::load: Loading service overrides from "./services/admiral-podrington/dev-ops.yml"
shipcat::diff: Writing crd for admiral-podrington to ./admiral-podrington.shipcat.gen.yml
shipcat::kube: kubectl diff -n=dev -f=./admiral-podrington.shipcat.gen.yml
```

then prints `kubectl diff` output (currently uses default `diff -u` output).

janslow:
I'm getting an error when running it in a clean checkout:

```bash
> cargo run -- diff profiles --git                                                                                                                                                                                           (dev-uk)
    Finished dev [unoptimized + debuginfo] target(s) in 1.01s
     Running `target/debug/shipcat diff profiles --git`
No stash entries found.
shipcat: diff error: Subprocess failure from git: 1
```

janslow:
Can we set the exit status to a non-zero value if there's a difference? (like `diff` and `git diff`)

shipcat_cli/src/diff.rs
```diff
@@ -0,0 +1,119 @@
...
+
+fn print_difference(lhs: &str, rhs: &str) -> Result<()> {
+    let Changeset { diffs, .. } = Changeset::new(lhs, rhs, "\n");
+    let mut t = term::stdout().unwrap();
+    // github style example via: https://github.com/johannhof/difference.rs/tree/master/examples
+    for i in 0..diffs.len() {
+        match diffs[i] {
+            Difference::Same(ref x) => {
+                t.reset().unwrap();
+                let _ = writeln!(t, " {}", x);
```
janslow: The indentation needs to be applied per line, as some diffs appear to be multiline.

image

Also, is it possible to reduce the amount of unchanged text that is shown? e.g., only show the changes and a few lines of context (like `git diff`/`diff`)?

shipcat_cli/src/diff.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Should be pretty safe. Stashes existing work, checks out master, compares,
+/// then goes back to previous branch and pops the stash.
+///
+/// Because this does fiddle with git state while running it is not the default implementation.
+pub fn file_vs_git(svc: &str, conf: &Config, region: &Region) -> Result<()> {
+    let aftermf = shipcat_filebacked::load_manifest(&svc, conf, region)?;
+    let after = serde_yaml::to_string(&aftermf)?;
+
+    // idea for actually getting before:
+    git(&["checkout", "master", "--quiet"])?;
```
gleaver: Perhaps `git checkout $(git merge-base master) --quiet` is more accurate?

shipcat_cli/src/diff.rs
```diff
@@ -0,0 +1,119 @@
...
+    // TODO: warn if branch is out of date
+    git(&["stash", "--quiet"])?;
+
+    // get old state
+    let beforemf = shipcat_filebacked::load_manifest(&svc, conf, region)?;
+    let before = serde_yaml::to_string(&beforemf)?;
+
+    // cleanup
+    git(&["stash", "pop", "--quiet"])?;
+    git(&["checkout", "-", "--quiet"])?;
```
gleaver: For things like this, I've been grabbing `git rev-parse --abbrev-ref HEAD` before moving HEAD in order to return the correct place

shipcat_cli/src/diff.rs
```diff
@@ -0,0 +1,119 @@
...
+                                    t.reset().unwrap();
+                                    let _ = write!(t, " ");
+                                }
+                                _ => (),
+                            }
+                        }
+                        let _ = writeln!(t, "");
+                    }
+                    _ => {
+                        t.fg(term::color::BRIGHT_GREEN).unwrap();
```
gleaver: what changes end up `REALLY_REALLY_GREEN`? Things which are more added than added?

shipcat_cli/src/diff.rs
clux: the difference crate doesn't have a good way of handling context. afaikt it's either no context or full context, because it all goes into the `Same()` blob. maybe we could figure out if the last one was a removal or addition and then either take the head or tail in next iteration though..

clux:
Thanks Jay, forgot that git stash fails to stash anything if there's nothing to stash :D

shipcat_cli/src/diff.rs
clux: honestly this was a complete copy paste from https://github.com/johannhof/difference.rs/blob/master/examples/github-style.rs

it looks nice, it's probably highlighting words like the newer git diff features.

janslow: Yep 

image

shipcat_cli/src/diff.rs
clux: ended up using shelling out to `diff(1)` instead for this

shipcat_cli/src/diff.rs
clux: `git merge-base master` returns 1 without printing anything if you're at master

shipcat_cli/src/diff.rs
clux: is this because we worried about `git checkout` race conditions here?
`git rev-parse --abbrev-ref HEAD` seems to always say master for me even if i'm on a branch

shipcat_cli/src/diff.rs
clux: ended up not using diff, and used `diff(1)` instead. at least that is predictable. tried a few other diffing libraries, none were great.

PR #138, opened 2019-03-27
[IDM-104] Add new authorization model

### LINKED ISSUE #123: New model for service authorization
We need to fully support JWT authorization from Shipcat instead of the Kong OAuth 2 plugin.

This is going to be using multiple plugins, the [Kong JWT plugin](https://docs.konghq.com/hub/kong-inc/jwt/) for basic validation (signature/expiry) and a custom plugin to perform additional validation (Audience, scope and region validation).

The `authorization` configuration will be part of the Kong configuration, but could be added to other service proxies (e.g., Envoy in a service mesh or Gate)

# Examples
Some examples of common use cases for services.

## Global/regional defaults
```yaml
# shipcat.conf
defaults:
    kong:
        # Secure by default
        authorization:
            enabled: true
            configuration: "{{ base_urls.services }}/.well-known/openid-configuration"
            legacy: false
            allowed_audiences:
            - https://babylonhealth.com/
            allow_anonymous: false
            remove_invalid_tokens: true
regions:
- name: ops-cluster
  kong:
    auth: none
    authorization:
      enabled: false
- name: dev-uk
  kong:
    auth: enabled
# prod-uk hasn't been migrated to use the new authorization, so uses oauth2 instead
- name: prod-uk
  kong:
    auth: oauth2
    authorization:
        enabled: false
    
```

## Simple service
```yaml
# shipcat.yml
kong:
    uri: ...
    ## The defaults are fine

# staging.yml
kong:
    authorization:
        legacy: true
```

Most services should need no custom configuration. Invalid/anonymous requests won't reach them.

Older versions of the service may not understand JWTs (e.g., the one deployed to `staging`), so will require the `legacy` mode.

## Service with anonymous access
```yaml
# shipcat.yml
kong:
    authorization:
        allow_anonymous: true
```

Some services (e.g., privacy service) will accept requests with no `Authorization` token and will add an `Babylon-Anonymous: true` header.

Requests with invalid/expired access tokens will have the `Authorization` header removed, to prevent programming bugs in the services from causing invalid tokens to be trusted.

## Service allowing non-standard tokens
```yaml
# shipcat.yml
kong:
    authorization:
        allow_anonymous: true
        remove_invalid_tokens: false
```

Some services may receive non standard tokens/`Authorization` headers (e.g., if they receive Slack webhooks). With the above configuration, the `Authorization` header will not be removed (but an `Babylon-Anonymous: true` will be added).

## Services with specific scope
```yaml
# shipcat.yml
kong:
    authorization:
        required_scopes: [clinical]
```

## Services supporting cookies
```yaml
# shipcat.yml
kong:
    authorization:
        cookies: true
```

Services which are called by web applications may recieve the access token in cookies.

With the above configuration, the access token will be extracted from the `Authorization` header (if present) and CSRF protection will be enabled.

## Services with no authorization
```yaml
# shipcat.yml
kong:
    authorization:
        enabled: false
```

A service may want to fully disable Kong-level authorization. This should be quite rare, with `allow_anonymous: true` covering most use cases for `auth: none` (without the risk of programming errors introducing security bugs).

# Structure
On a resolved manifest, the schema will be as follows:

```rust
pub struct Kong {
    /// ...

    /// If present, Kong is configured to perform authorization for the service
    /// If omitted, no authorization is performed
    pub authorization: Option<Authorization>,
}

pub struct Authorization {
    /// OpenID Connect Discovery URL (for discovery of JWT public keys from the JWKS URL)
    ///
    /// e.g., https://login.microsoftonline.com/babylonpartners.onmicrosoft.com/v2.0/.well-known/openid-configuration or https://janslow-babylon.eu.auth0.com/.well-known/openid-configuration
    pub configuration: String,
    
    /// Add legacy headers
    ///
    /// Add Kong OAuth 2.0 style legacy headers, e.g., `X-Authenticated-Userid` for services which aren't aware of JWTs
    pub legacy: bool,

    /// Allowed values for the `aud` claim of the JWT payload.
    pub allowed_audiences: Vec<String>,

    /// Are anonymous requests allowed to reach the service?
    ///
    /// If true, requests with no `Authorization` header (or an invalid/expired JWT) will be proxied to the service (but will receive an `Anonymous: true` header)
    /// If false, they will be rejected (with a 401 response)
    pub allow_anonymous: bool,
    
    /// Should invalid/expired tokens be stripped from the upstream request?
    ///
    /// If true, invalid `Authorization` headers will be removed from the request.
    pub remove_invalid_tokens: bool,

    /// What JWT scopes are required for the service?
    /// 
    /// If the JWT does not contain the required scopes, the request will be rejected with a 401.
    pub required_scopes: Vec<String>,

    /// Are tokens in cookies supported?
    ///
    /// If true, CSRF protection is enabled and access tokens are extracted from cookies.
    pub cookies: bool,
}
```

This would be converted into configuration for the service proxy (currently Kongfig plugin entries).

Assuming #120 is done, this would be represented using the following source struct.

```rust
pub struct AuthorizationSource {
    /// Allows easy enabling/disabling at different levels.
    pub enabled: Option<bool>,
    pub configuration: Option<TemplatedString>,
    pub legacy: Option<bool>,
    /// Must be non-empty
    pub allowed_audiences: Option<Vec<String>>,
    pub allow_anonymous: Option<bool>,
    pub remove_invalid_tokens: Option<bool>,
    pub required_scopes: Option<Vec<String>>,
    pub cookies: bool,
}

/// Used in shipcat.conf
pub struct KongDefaults {
    pub auth: Option<Authentication>

    pub authorization: Option<AuthorizationSource>,
}

pub struct KongSource {
    /// ...

    /// If set explicitly, treated as `authorization.enabled: false`
    /// Post migration, this will be removed
    pub auth: Option<Authentication>

    pub authorization: Option<AuthorizationSource>,
}
```

# Possible Future Enhancements
* Authorization Permissions
    * Services could define the permissions they may require (which will be used to configuration the "authorization service")

janslow:
Introduces the MVP of the `Authorization` struct, used to configure Kong authz instead of the old properties like `auth` and `oauth2_anonymous`.

This is a backwards compatible change. I've diffed output of the latest Shipcat with this one in dev and prod, the only difference is with `shipcat kong`:

```diff
@@ -5937,6 +6145,8 @@ apis:
             accept_http_if_already_terminated: false
       - name: jwt
         ensure: removed
+      - name: jwt-validator
+        ensure: removed
     attributes:
       hosts: {}
       uris:
```

(For each API, but `jwt-validator` is a new Kong plugin so is unused)

Fixes #123


shipcat_definitions/src/region.rs
```diff
@@ -491,7 +509,10 @@ pub struct Region {
     /// All webhooks
     pub webhooks: Option<Vec<Webhook>>,
     /// CRD tuning
-    pub customResources: Option<CRSettings>
+    pub customResources: Option<CRSettings>,
+    /// Default values for services
+    #[serde(skip_serializing, default)]
+    pub defaults: DefaultConfig,
```
janslow: If a `shipcat_filebacked::ConfigSource` struct was split from `Config`, `RegionConfig#defaults` and `Config#defaults` could be of type `ManifestDefaults` instead, making it easier to set global/regional default values.

shipcat_definitions/src/structs/kong.rs
```diff
@@ -22,10 +22,6 @@ pub struct Kong {
     #[serde(default)]
     pub upstream_url: String,
 
-    /// Whether the oauth2 plugin will be applied or not to this api
-    #[serde(default, skip_serializing)]
-    pub unauthenticated: bool,
```
janslow: `auth: AuthType::None` is equivalent to `unauthenticate: true`, so this only exists in `KongSource` now.

shipcat_definitions/src/structs/kongfig.rs
```diff
@@ -345,29 +355,51 @@ pub fn kongfig_apis(from: BTreeMap<String, Kong>, config: KongConfig) -> Vec<Api
...
+            if a.allow_cookies {
+                plugins.push(ApiPlugin::JsonCookiesToHeaders(PluginBase::default()));
+                plugins.push(ApiPlugin::JsonCookiesCsrf(PluginBase::default()));
+            } else {
+                plugins.push(ApiPlugin::JsonCookiesToHeaders(PluginBase::removed()));
+                plugins.push(ApiPlugin::JsonCookiesCsrf(PluginBase::removed()));
+            }
+        } else {
+            // OAuth2 plugins
+            plugins.push(ApiPlugin::Oauth2(match v.auth {
```
janslow: After [IDM-105](https://babylonpartners.atlassian.net/browse/IDM-105) (when all services use `authorization` instead of the old `auth` properties, this `else` branch can be deleted.

shipcat_filebacked/src/authorization.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+impl AuthorizationSource {
+    pub fn build(self) -> Result<Option<Authorization>> {
+        if !self.enabled.unwrap_or_default() {
+            return Ok(None);
+        }
+        let allowed_audiences = self.allowed_audiences.unwrap_or_default();
+        if allowed_audiences.is_empty() {
+            bail!("allowed_audiences must contain at least one audience");
```
janslow: I'd like to find a nicer way to declare validation.

I tried https://crates.io/crates/validator briefly, but it couldn't handle the `Option`s

clux: hm, there is an issue on it in https://github.com/Keats/validator/issues/64 - but supposedly it should work

shipcat_filebacked/src/kong.rs
```diff
@@ -0,0 +1,152 @@
...
+            }
+            // authorization is enabled
+            (None, _, Some(a)) | (Some(Authentication::Jwt), _, Some(a)) => {
+                Ok((Authentication::Jwt, Some(a)))
+            }
+            (Some(_), _, Some(_)) => bail!("auth must be unset or JWT if authorization is enabled"),
+            // otherwise
+            (Some(x), _, _) => Ok((x.clone(), None)),
+            (None, _, _) => Ok((Authentication::default(), None)),
+        }
```
clux: beautiful

shipcat_filebacked/src/authorization.rs
clux: btw, if even the source needs to have one element in this vec, should it be Option'd?

shipcat_filebacked/src/kong.rs
```diff
@@ -0,0 +1,152 @@
...
+            if self.cookie_auth_csrf.is_some() {
+                bail!("cookie_auth_csrf and authorization properties are mutually exclusive")
+            }
+            if self.oauth2_anonymous.is_some() {
+                bail!("oauth2_anonymous and authorization properties are mutually exclusive")
+            }
+            if self.oauth2_extension_plugin.is_some() {
+                bail!("oauth2_extension_plugin and authorization properties are mutually exclusive")
+            }
+        }
```
clux: :+1:

shipcat_filebacked/src/authorization.rs
janslow: It's optional in any individual file, but only required in an assembled manifest.

In this case, it's probably going to be set in `shipcat.conf` as a regional/global default, and only overridden on a couple of specific services.

I'd like to pull more of the `Manifest#verify` into this logic, so I'll have a better look at some validation crates then.

shipcat_filebacked/src/kong.rs
janslow: I'd forgotten how great pattern matching was because of too long working with Java ❤️

PR #139, opened 2019-03-29
Support team namespaces for kubectl

gahill:


examples/charts/base/templates/configmap.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ .Values.name }}-config
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ .Values.namespace }}
```
clux: btw

```
    release: {{ .Release.Name }}
    heritage: {{ .Release.Service }}
```

will be:

```
    release: RELEASE-NAME
    heritage: Tiller
```
when not upgraded through helm. This is probably fine, even if it's false initially.

examples/shipcat.conf
```diff
@@ -24,6 +24,12 @@ github:
 
 teams:
 - name: Doves
+  namespace: doves
+  owners:
+  - name: Eirik Albrigtsen
+    github: clux
+    slack: "@clux"
+- name: TheOthers
```
clux: i am in both!? :laughing:

shipcat_definitions/src/config.rs
```diff
@@ -93,6 +98,16 @@ impl Team {
...
             bail!("Every team must have a default notifications channel declared");
         }
+
+        if let Some(namespace) = &self.namespace {
+            use regex::Regex;
+            let re = Regex::new(r"^[a-z\-]{1,30}$").unwrap();
+
+            if !re.is_match(&namespace) {
+                bail!("Valid namespace are lower case alpha with dashes only.")
+            }
```
clux: we need a blacklist of namespaces from somewhere `["kube-system", "monitoring", "flux", ...]` should probably be a property on `Region`

shipcat_filebacked/src/manifest.rs
```diff
@@ -157,15 +158,32 @@ impl ManifestSource {
...
     }
 
+    /// Coalesce the namespace from team and region definitions in Shipcat.conf
+    /// assumes the region will always have a namespace to fallback on
+    /// 
+    /// region namespace set | team exists | team namespace set | output
+    /// ---------------------+-------------+--------------------+---------------
+    /// yes                  | yes         | yes                | team namespace
+    /// yes                  | yes         | no                 | region namespace
+    /// yes                  | no          | no                 | region namespace
```
clux: generally team's existence is verified by validate, but this is safe

clux:
looks sensible. it might lack some changes:

- kong's upstream_uri typically go to `/{svc}.{ns}.svc.cluster.local` not sure if `{ns}` therein is picked up from manifest or from `Region`
- Region.namespace uses elsewhere ought to be reviewed

clux:
so reconcile under this model just works presumably? initially duplicating deployments?

shipcat_definitions/src/config.rs
gahill: Will add this

gahill:
> looks sensible. it might lack some changes:
> 
> * kong's upstream_uri typically go to `/{svc}.{ns}.svc.cluster.local` not sure if `{ns}` therein is picked up from manifest or from `Region`
> * Region.namespace uses elsewhere ought to be reviewed

As discussed I will verify the usage of region vs manifest and standardise on manifest. Will add a flag to the region config so we can turn this on / off per cluster

shipcat_definitions/src/config.rs
gahill: Should this be at a region level. Would we want to allow kube-system for some regions?

clux: probably not, kube-system is not for user services

clux:
https://github.com/Babylonpartners/shipcat/issues/140 also worth tackling for the gc aspect of this

Issue #85, opened 2019-01-21
Regional rollouts


clux:
When doing larger changes to core or charts it would be good to have a regional rollout of the shipcat version, or the charts being used.

The charts would be pretty easy:

- define chart paths in `Region` configs: (say `{'base': 'charts/base', 'elb': 'charts/elb'}`)
- override in test regions

shipcat version would be a bit more complicated because you need to read the shipcat config before checking out the version of shipcat you need to roll things out.

We could establish a convention of putting the `shipcat` version in the `Region` struct, but that means you'd have to do a bit of an annoying `yq` on the file to figure out what version to use, and it would make it hard to change those structs at the same time without removing all the `deny_unknown_fields` annotations.

yq magic would be:

```sh
yq -y ".regions | map(select(.name == \"my-region\") .)" < ${SHIPCAT_MANIFEST_DIR}/shipcat.conf | yq '.version'
```

Which makes the unknown fields question the key one. We should probably change it to a `warn` to better deal with breaking changes for raftcat anyway.

Better suggestions welcome, but this probably solves it:

- [ ] put shipcat version to use per region in `Region` struct
- [ ] ci should use `yq` to decide shipcat version for reconciliation
- [ ] `s/deny/warn` in unknown_fields if possible for better backwards/forwards compat

JoseAlban:
Hmm - that's a good feat to have - but thinking of maybe having a global version and region-overrides, so shipcat itself can exec again into appropriate version if there's an override - this way, if eg v0.84.0 knows the region-override field, we don't need to deal with yq or unknown_fields compat (maybe?).

janslow:
Rather than rolling out per region, can we just restrict it to per environment. i.e., you can set `dev: 0.82.0` and `staging: 0.81.0` etc.

This will help to limit the complexity of manifests, especially as the number of regions grows.

clux:
We're experimenting with this now. A key subtlety:

client side validation in manifests repo needs to be parametrised per region/environment. otherwise we'll get reconciles failing for older versions when manifests were validated for a newer version.

@JoseAlban is having a go at it today.

clux:
More thoughts on this. A simplifying start is to just replace `version` with `versions` in the config:

```yaml
versions:
  dev: 0.88.0
  staging: 0.87.0
  preprod: 0.87.0
  prod: 0.87.0
```

so that:
- reconcile jobs can grab it from `yq ".versions.dev" -r < shipcat.conf`
[...]

PR #143, opened 2019-04-03
regional shipcat version rollout support

clux:
for #85


shipcat_definitions/src/config.rs
```diff
@@ -171,8 +171,8 @@ pub struct Config {
     /// Team definitions
     pub teams: Vec<Team>,
 
-    /// Shipcat version pin
-    pub version: Version,
```
janslow: Removing this field is going to break old Shipcats, without giving the old users a nice error message.

E.g., using Shipcat 0.95.0 to try to use this branch's `shipcat.conf`:
image

Breaking it is probably fine, but you may get a bump in "Shipcat isn't working" issue reports. Alternatively, leave `version: 0.96.0` around for a while (but unused).

clux: thanks, i've put it back with a serde default annotation.

PR #146, opened 2019-04-04
Kong struct cleanup

janslow:
Simplify some the fields of the Kong struct:
* `add_headers` is now `BTreeMap<String, String>` instead of `Option<BTreeMap<String, String>>`
* Remove `host` (it was only used for deserialization, which is now handled by `KongSource`)
* `hosts` is now a `Vec<String>` instead of comma separated `String`
  * Custom deserializer allows parsing `hosts` as a string or list from `shipcat.conf`
* `verify()` method was unused (but logic is now in `KongSource#build()`)


shipcat_definitions/src/structs/kong.rs
```diff
@@ -138,12 +130,47 @@ pub struct Kong {
...
     /// ```
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub add_headers: Option<BTreeMap<String, String>>,
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub add_headers: BTreeMap<String, String>,
 }
 
 fn preserve_host_default() -> bool { true }
 
+fn comma_separated_string<'de, D>(deserializer: D) -> Result<Vec<String>, D::Error>
```
clux: ah, and here i thought you might have an interesting migration on your hands

shipcat_filebacked/src/kong.rs
```diff
@@ -137,16 +130,18 @@ impl KongSource {
...
+            hosts.split(",").map(String::from).collect()
         } else {
-            Some(tophosts.join(","))
+            Default::default()
+        };
+        match (tophosts.as_slice(), &self.host, hosts.as_slice()) {
+            (_, None, []) => Ok(tophosts),
+            ([], None, _) => Ok(hosts),
+            ([], Some(host), []) => Ok(vec![format!("{}{}", host, region.kong.base_url)]),
+            (_, _, _) => bail!("hosts, kong.hosts and kong.host are mutually exclusive"),
```
clux: beautiful

Issue #142, opened 2019-04-03
team/env based slack notification control


clux:
`teams: ` struct in `shipcat.conf` should have more controls on where notifications go, perhaps an enum and a map in `config.rs`:

```rust
enum NotificationMode {
  NotifyMaintainers,
  MessageOnly,
  Silent,
}
```
then add a map of environments to setting to each `Team`:

```rust
slackSettings: BTreeMap<Environment, NotificationMode>
```

Once these structs exist, we can modify `slack.rs` to act on the team settings based on the environments.
The main `send()` function handles all slack events atm.

clux:
This was done in #149

PR #149, opened 2019-04-09
Issue #142 slack notification control

hughsimpson:


shipcat_definitions/src/config.rs
```diff
@@ -40,7 +40,13 @@ impl Default for ManifestDefaults {
     }
 }
 
-
+/// Modes for slack notification
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub enum NotificationMode {
+    NotifyMaintainers,
+    MessageOnly(Vec<Contact>),
```
clux: `MessageOnly` was intended to be a message without at'ing someone. Otherwise we'd have to duplicate contacts quite a lot. 

Similarly; `NotifyMaintainers` was meant to take `mf.metadata.contacts` so that we wouldn't have to have it in the config duplicated. But I think that's what it's doing currently.

shipcat_cli/src/slack.rs
```diff
@@ -194,8 +209,12 @@ fn create_github_compare_url(md: &Metadata, vers: (&str, &str)) -> SlackTextCont
     Link(SlackLink::new(&url, &short_ver(vers.1)))
 }
 
+fn contacts_to_text_content(contacts: &Vec<Contact>) -> Vec<SlackTextContent> {
+    contacts.iter().map(|cc| { User(SlackUserLink::new(&cc.slack)) }).collect()
+}
+
 fn infer_slack_notifies(md: &Metadata) -> Vec<SlackTextContent> {
-    md.contacts.iter().map(|cc| { User(SlackUserLink::new(&cc.slack)) }).collect()
+    contacts_to_text_content(&md.contacts)
```
clux: this function is basically useless now

shipcat_cli/src/slack.rs
```diff
@@ -137,13 +140,29 @@ fn send_internal(msg: Message, chan: String) -> Result<()> {
         texts.push(infer_ci_links());
     }
 
+    let notificationMode: &NotificationMode = if_chain!{
+        if let Some(ref md) = msg.metadata;
+        if let Some(team) = conf.teams.iter().find(|t| t.name == md.team );
+        if let Some(slackSettings) = team.slackSettings.get(&env);
```
clux: ah that's why you added the crate. seems reasonable.

shipcat_cli/src/slack.rs
```diff
@@ -137,13 +140,29 @@ fn send_internal(msg: Message, chan: String) -> Result<()> {
...
         if !msg.quiet {
-            texts.push(Text("<- ".to_string().into()));
-            texts.extend(infer_slack_notifies(md));
+            match notificationMode {
+                NotificationMode::NotifyMaintainers => {
+                    texts.push(Text("<- ".to_string().into()));
+                    texts.extend(contacts_to_text_content(&md.contacts));
+                }
+                default => {},
+            }
```
clux: looks pretty clean. you can do `_ => {}` instead of the explicit default.

clux:
This looks good to me if it works. Needs to test on an up to date branch and then version bump with `./scripts/bump_version.sh 0.107.0 0.108.0` before making a release.

shipcat_cli/src/slack.rs
clux: lol the if_chain crate is some black magic: https://github.com/lfairy/if_chain/blob/master/src/lib.rs#L164-L238

shipcat_definitions/src/config.rs
```diff
@@ -40,7 +40,13 @@ impl Default for ManifestDefaults {
     }
 }
 
-
+/// Modes for slack notification
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+pub enum NotificationMode {
```
janslow: @clux, what should be the casing for serialized enums? We've currently got a mix of lowercase, kebab-case, UPPERCASE and SCREAMING_SNAKE_CASE, but we should try to standardise on one. (my preference is anything except lowercase/UPPERCASE)

hughsimpson: There's also a bit of camel case already (versioningScheme, ConfigType). I think at the moment the kebab-case is just for instances where the second part 'qualifies' the first? (e.g. `preprod-uk`, `preproduk-green`). IMO reserving kebab for those cases makes sense

clux: PascalCase is the default from rust that we tend to use. But there's like one place that has UPPERCASE (protocol name), and a few with lowercase (like kubernetes verbs). Single-word lowercase is quite common. I'm not sure there's always a good answer here. I like to not have special annotations more than necessary, so lean towards `PascalCase`.

PR #156, opened 2019-04-11
crd reconciliation for #150

### LINKED ISSUE #150: Smarter reconciles: Part 1: CRD configuration
Since kube 1.12 we now get proper output from crd application:

```
11:23:29 shipcatmanifest.babylontech.co.uk/changed-service-name configured
11:23:29 shipcatmanifest.babylontech.co.uk/unchanged-service-name unchanged
```

## Plan

- Use `configured/unchanged` output from `cluster crd reconcile` to decide if we should also `helm upgrade`
- Add a `ReconciliationMode::CrdBurrowed` enum as configurable on `Region` (non-default) 
- Gradually roll out this mode to all clusters as we stop running `cluster helm reconcile`
- Deprecate the default `ReconciliationMode` (swap defaults to `CrdBurrowed`)
- Deprecate `cluster helm reconcile`

Requires kubernernetes 1.12 to tell if it was changed or not.  Note that still uses tiller, is just is a lot more selective with when it calls `helm upgrade` and `helm diff upgrade` (only when the CRD was `configured).

## Pros:

- much faster reconciles 99% of services generally do not change  every run (should expect an order of magnitude speedup)
- less load on vault (won't fetch all secrets)
- less load on tiller/etcd (won't need to diff on 99% of services) 
- can be gradually rolled out to clusters behind an enum
- provides a clear migration paths away from tiller when we can start using ownerRefs of crds in #140 without writing another tiller clone

## Cons

- vault changes won't be automatically propagated (will need a manual push or a [custom vault webhook plugin](https://github.com/idcmp/vault-plugin-secrets-webhook))

Secret existence can still be checked externally on validated merge to a manifest repo using `shipcat secret verify-region`, but it's a cheaper and slightly inaccurate operation (LIST with a set diff rather than a full GET on all secrets).

This is part one of a migration away from tiller. See #151 for part two.

clux:
when a region has set `reconciliationMode: CrdBurrowed`, we will reconcile from crd changes only. This closes #150


clux:
This works now. Behind a regional feature flag. Bumped to 0.100.0.

shipcat_cli/src/kube.rs
```diff
@@ -451,9 +452,20 @@ pub fn apply_crd<T: Into<Crd<T>> + Serialize>(name: &str, data: T, ns: &str) ->
...
+    let (out, status) = kout(applyargs.clone())?;
+    debug!("{}", out); // always print kube output from this
+    if !status {
+        bail!("subprocess failure from kubectl: {:?}", applyargs);
+    }
+    let changed = if out.contains(&format!("{} configured", name)) {
+        true
+    } else if out.contains(&format!("{} created", name)) {
+        true
+    } else {
```
gleaver: Explicitly catch known values and warn/error on unexpected strings

shipcat_cli/src/cluster.rs
```diff
@@ -101,7 +101,16 @@ fn crd_reconcile(svcs: Vec<SimpleManifest>, config: &Config, region: &Region, n_
 
 fn crd_reconcile_worker(svc: &str, conf: &Config, reg: &Region) -> Result<()> {
     let mf = shipcat_filebacked::load_manifest(svc, conf, reg)?;
-    kube::apply_crd(svc, mf, &reg.namespace)?;
+    if kube::apply_crd(svc, mf.clone(), &reg.namespace)? {
```
gleaver: Add some kind of envvar/flag override for this to allow full reconcile if needed.

shipcat_definitions/src/region.rs
```diff
@@ -456,6 +456,35 @@ impl ToString for Environment {
...
+pub enum ReconciliationMode {
+    /// Tiller owned, apply every time
+    ///
+    /// If anything's different in diff (even secrets), then helm apply.
+    MassHelm,
+
+    /// Tiller owned, CRD based decision
+    ///
+    /// If CRD was configured, then helm apply.
+    CrdBurrowed,
```
gleaver: `Borrowed`?

shipcat_cli/src/cluster.rs
clux: added a SHIPCAT_MASS_RECONCILE evar

shipcat_definitions/src/region.rs
clux: typoed

PR #163, opened 2019-04-23
Allow custom metadata properties to be defined

### LINKED ISSUE #157: Allow an `special` key for non-typechecked properties?
An idea. Can potentially be used to make developing against non-standard charts easier.

e.g.

```yaml
name: somesvc
special:
  anything: goes
  here: 2
```

See [std::any](https://doc.rust-lang.org/std/any/trait.Any.html) might work, but don't think you can serialize/deserialize it. Might have to tell serde it's a json::Value or something similar.

Thoughts?

janslow:
Custom metadata properties can be defined in `shipcat.conf`, so that they will be accepted (and available via the CRD).

Fixes #157


janslow:
See https://github.com/Babylonpartners/shipcat/pull/163 as an example (it should fail on in it's current state, but isn't due to #162)

shipcat_definitions/src/structs/metadata.rs
```diff
@@ -107,6 +108,10 @@ pub struct Metadata {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub docs: Option<String>,
     // TODO: generate swagger docs url from region and service name
+
+    /// Custom metadata, keys defi
```
clux: abbreviated word

shipcat_definitions/src/structs/metadata.rs
```diff
@@ -107,6 +108,10 @@ pub struct Metadata {
...
+    #[serde(flatten)]
+    pub custom: BTreeMap<String, String>,
```
clux: good place to put it :+1:

shipcat_definitions/src/structs/metadata.rs
```diff
@@ -151,6 +156,11 @@ impl Metadata {
                 bail!("Runbook must be in markdown or restructured text in the service repo");
             }
         }
+        for k in self.custom.keys() {
+            if !allowedCustomMetadata.contains(k) {
+                bail!("{} is not an allowed metadata property", k);
+            }
+        }
```
clux: nice and clean.

shipcat_definitions/src/structs/metadata.rs
janslow:
```suggestion
    /// Custom metadata, keys defined in shipcat.conf
```

I got distracted part way through.

PR #164, opened 2019-04-24
Fix empty hosts property in Kongfig output

janslow:
Kong represents an empty list as `{}`, so Kongfig expects the same to correctly diff the state to work out required changes.

Should fix the performance issue where kongfig tries to update an API which is already up to date.


clux:
Looks sane, presumably you've tested the output.

janslow:
Yep, diffed them, it's just lots of this.

```diff
@@ -12938,7 +12938,7 @@ apis:
       - name: jwt-validator
         ensure: removed
     attributes:
-      hosts: []
+      hosts: {}
       uris:
         - /
       methods: {}
```

I've confirmed that syntax works on my local Kong, but it only undoes the behaviour introduced by https://github.com/Babylonpartners/shipcat/pull/146

This release can be rolled out per region though as well.

PR #177, opened 2019-05-07
dependency bump + duplicate includes + shipcat env

clux:
only important thing here is the removal of `region_name` in  templating, which is replaceable with `region` in a pr to manifests.


shipcat_definitions/src/template.rs
```diff
@@ -76,7 +76,6 @@ impl Manifest {
         ctx.insert("kong", &reg.kong);
         ctx.insert("cluster", &reg.cluster.clone());
         ctx.insert("namespace", &reg.namespace.clone());
-        ctx.insert("region_name", &reg.name.clone());
```
janslow: We already have a `region` property equal to this.

clux: exactly :slightly_smiling_face:

janslow: 🤦‍♂

Issue #184, opened 2019-05-09
Codeowners not generated for team with only githubAdmin


janslow:
If you have the following setup:

```yaml
# shipcat.conf
teams:
- name: my-team
  githubAdmin: my-team-admins

# services/my-service/shipcat.yml
metadata:
  team: my-team
```

I'd expect `CODEOWNERS` to contain the following:

```
# CODEOWNERS
services/my-service/* @Organisation/my-team-admins
```

However, the entry for the service is only generated when there is at least one entry in the team's `owners` list.

clux:
ooh wow, good catch.

PR #189, opened 2019-05-13
codeowners tweaks

clux:
for #184 but also standardising output a bit better.


shipcat_cli/src/get.rs
```diff
@@ -51,15 +51,13 @@ pub fn codeowners(conf: &Config) -> Result<Vec<String>> {
         // teams guaranteed by validates on Manifest and Config
         if let Some(t) = &conf.teams.iter().find(|t| t.name == md.team) {
             if let Some(gha) = &t.githubAdmins {
-                ghids.push(format!("@{}/{}", org, gha));
+                ghids.push(format!("@{}/{}", org.to_lowercase(), gha));
```
janslow: Should `gha` be lowercased as well?

shipcat_cli/src/get.rs
```diff
@@ -51,15 +51,13 @@ pub fn codeowners(conf: &Config) -> Result<Vec<String>> {
...
             }
             // TODO: take this out now that we have teams?
             for o in t.owners.clone() {
                 ghids.push(format!("@{}", o.github.unwrap()));
             }
-            if !t.owners.is_empty() {
-                output.push(format!("services/{}/* {}", mf.name, ghids.join(" ")));
-            }
+            output.push(format!("services/{}/* {}", mf.name, ghids.join(" ")).trim().into());
```
janslow: What if `ghids` is empty (e.g., `owners` is empty and `githubAdmins` is none)?

shipcat_cli/src/get.rs
clux: that's already enforced due to vault policy reconcile

shipcat_cli/src/get.rs
clux: have fixed that now.

PR #190, opened 2019-05-14
[ENC-34] Configure Kong to add Upstream-Service header to requests

janslow:
Configure all Kong APIs associated with a service to use the [Request Transformer plugin](https://docs.konghq.com/hub/kong-inc/request-transformer/) and configure it to set the `Upstream-Service: $service_name` header.

This allows upstream proxies (e.g., Envoy/Istio) to route requests to the correct service based on that header.


clux:
I see you're taking out a lot of `deny_unknown_fields` attrs. Does this not apply to these structs?

janslow:
I'm also removing all of the `Deserialize` traits from the Kongfig fields, as they're unused

PR #194, opened 2019-05-15
Adding pod-annotations to shipcat

atamrat-babylon:
Need the ability to add pod annotations to deployments, cronJobs, etc. to enable tools like kube2iam.

Not sure if this kind of functionality should have tests added to it, since the most logic is on the helm templating side of things. Tried to copy the pattern for `labels` and `service_annotations`. 

Let me know if any other changes are needed


examples/charts/base/templates/cronjobs.yaml
```diff
@@ -18,6 +18,11 @@ spec:
   concurrencyPolicy: Forbid
   schedule: "{{ $v.schedule }}"
   jobTemplate:
+{{- if $.Values.podAnnotations }}
```
janslow: I wonder whether it'd be better to set `podAnnotations` on the individual pod, instead of the whole Shipcat service.

E.g., allow it to be set on a specific CronJob instead.

atamrat-babylon: If i understand you correctly, i agree. Great catch! i forgot that cron jobs have a template for the job WHICH ALSO HAS a template for the pod. @janslow

janslow: e.g.,
```yaml
# services/my-service/shipcat.yml
podAnnotations:
  foo: 'main'
workers:
- name: my-worker
  podAnnotations:
    foo: worker
cronJobs:
- name: my-cron
  podAnnotations:
    foo: scheduled
```

atamrat-babylon: @janslow I went ahead and updated per your suggestion, could you double check to make sure it makes sense?

PR #195, opened 2019-05-20
[DAT-10] Add ZK ensemble support in Kafka struct

frenayj:
image


shipcat_definitions/src/region.rs
```diff
@@ -116,6 +116,11 @@ pub struct KafkaConfig {
     /// These are injected in to the manifest.kafka struct if it's set.
     pub brokers: Vec<String>,
 
+    /// Zookeeper urls in "hostname:port" format.
+    ///
+    /// These are injected in to the manifest.kafka struct if it's set.
```
clux:
```suggestion
    /// These are injected in to the manifest.kafka struct if it's set.
    #[serde(default)]
```

clux:
premature tick, do add the default annotation first. and also run:

```sh
./scrips/bump_version.sh 0.106.2 0.107.0
```

then git commit the changes.

clux:
on merge, make the `0.107.0` release on github.

PR #219, opened 2019-07-06
no more special casing vaults..

clux:


clux:
someone plz bump and propagate this raftcat version :-)

PR #220, opened 2019-07-10
[CRP-3069] Add thresholds to kubernets probes

riman-babylon:


shipcat_definitions/src/structs/probes.rs
```diff
@@ -60,12 +60,23 @@ pub struct Probe {
...
+    pub failureThreshold: u32,
 }
 
 // 5 is kube standard delay default, we set it a little higher
 fn initial_delay_seconds_default() -> u32 { 30 }
 // how frequently to poll
 fn period_seconds_default() -> u32 { 5 }
+// Default values from Kubernetes
+fn success_threshold_default() -> u32 { 1 }
+fn failure_threshold_default() -> u32 { 3 }
```
clux: it might be better to have `Option<u32>` and skip serializing if empty. This would force us to use kube defaults without hardcoding what they are (since that might change).

PR #221, opened 2019-07-11
Might be worth mentioning SHIPCAT_MANIFEST_DIR

AndrewGHC:


janslow:
It's not required (as it defaults to your current dir), but it may be useful to document it somewhere.

clux:
It is mentioned in the `building.md` page, but it would be good if the homebrew formula suggested adding it to your `~/.bash_profile` on install (or maybe just doing it for them).

PR #222, opened 2019-07-19
EKS auth command

RhysStansfield:
Add `shipcat auth` command.

Reimplement's the `eks-auth.sh` script from manifests repo in shipcat.

Respects -r flag or defaults to current kube environment context.

Example:
```bash
$ shipcat auth # will default to whatever kube environment context kube is currently set
$ shipcat auth -r dev-uk # will auth in dev-uk
```

TODO:
* Can we handle firing off teleport request via curl?  Or auto-closing of browser window?


shipcat_cli/src/kube.rs
```diff
@@ -46,6 +46,29 @@ pub fn current_context() -> Result<String> {
     Ok(res)
 }
 
+pub fn set_context(context: &String, args: Option<Vec<String>>) -> Result<String> {
+    let mut arg_list = vec!["config".into(), "set-context".into(), context.into()];
+    if let Some(mut args) = args {
```
clux: rather than having optional args, just pass in the args as a vector, then iterate over the potentially empty one

clux: rather than having optional wrapped vec of args, just pass in the args as a potentially empty vector, then iterate over it. this is what we do elsewhere

shipcat_cli/src/kube.rs
```diff
@@ -46,6 +46,29 @@ pub fn current_context() -> Result<String> {
     Ok(res)
 }
 
+pub fn set_context(context: &String, args: Option<Vec<String>>) -> Result<String> {
```
clux: &str and Vec<String> are better types here

shipcat_cli/src/auth.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+fn check_cluster_matches_region(cluster: &Cluster, region_name: &String) -> bool {
+    cluster.regions.iter().any(|region_str| region_str == region_name)
+}
+
+pub fn eks_auth(conf: &Config, region: &Region) -> Result<()> {
+    use std::process::Command;
+    let s = Command::new("which").args(vec!["tsh"]).output()?;
+    let out : String = String::from_utf8_lossy(&s.stdout).into();
```
clux: : String here is unnecessary since you're doing `String::from_something` which obviously returns String :-)

shipcat_cli/src/auth.rs
```diff
@@ -0,0 +1,40 @@
...
+
+fn extract_config(cluster_pair: (&String, &Cluster)) -> Option<Cluster> {
+    Some(cluster_pair.1.clone())
+}
+
+fn find_cluster(clusters: &BTreeMap<String, Cluster>, region_name: &String) -> Option<Cluster> {
+    match clusters.iter().find(|&cluster_tuple| check_cluster_matches_region(&cluster_tuple.1, &region_name)) {
+        Some(cluster) => extract_config(cluster),
+        None => None,
+    }
```
clux: a straight loop here would probably be more readable, it's a lot of magic tuples and find into ad-hoc functions atm.

```rust
for (_, c) in clusters {
    if c.regions.iter().any(|r| r == region_name) {
        return Some(c.clone()) }
    }
    None
}
```

RhysStansfield: Much better 👌

shipcat_cli/src/kube.rs
RhysStansfield: Agreed, wasn't sure of this - thanks for clarifying

shipcat_cli/src/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+        if let Some(teleport) = &cluster.teleport {
+            info!("Authenticating with {}", teleport);
+            let proxy_str = format!("--proxy={url}:443", url = &teleport);
+            Command::new("tsh").args(vec!["login", &proxy_str, "--auth=github"]).output()?;
+            // NB: tsh creates a cluster entry in ~/.kube/config named after the url
+            // We cannot customize this name the name of this cluster
+            let args = vec![
+                format!("--cluster={}", &teleport),
+                format!("--user={}", &teleport),
+                format!("--namespace={}", region.namespace),
```
clux: made a small tweak here to not quote the args (was actually broken, but you might not have used the new context you created so it MIGHT have worked for you.

also added namespace flag which was needed.

shipcat_cli/src/auth.rs
```diff
@@ -0,0 +1,46 @@
...
+            ];
+            kube::set_context(&region.name, args)?;
+            kube::use_context(&region.name)?;
+        } else {
+            bail!("Cluster for {} does not have a teleport url", region.name);
+        }
+    } else {
+        bail!("Region {} does not have a cluster", region.name);
+    }
```
clux: also some more error handling at the end

clux:
And tried a bunch to maintain the existing connections, which was a pain because tsh does not seem to have a nice way of getting the status of an existing cluster. See https://github.com/gravitational/teleport/issues/2869

Ended up doing the equivalent of a tsh status with a double grep:

```
tsh status | grep "teleport-url" -A 3 |grep EXPIRED
```

PR #225, opened 2019-07-24
Add `Apps` API Group.

carlosgaldino:


clux:
LGTM but bump with `scripts/bump_version.sh current newver` all versions stay in sync

PR #226, opened 2019-07-26
Remove `crd` compile feature

janslow:
Can't quite get rid of all the conditional compilation, as there's still a lot of conditional deserialization in `shipcat.conf`.

However, there's no longer any references to the `crd` feature, so that can go 🎉  (which also means we can test raftcat in CI).


janslow:
I've diffed the output of `shipcat kong`, `shipcat crd`, `shipcat validate` and `shipcat helm` for each service in `dev-uk`, and there are no differences so far.

PR #227, opened 2019-07-26
Rewrite shipcat apply

clux:
To prepare for helm upgrade -> kubectl apply change.

This also does a whole bunch of simplifying changes without changing most of the entrypoints yet.

The idea is to remove most of the helm module in shipcat_cli, create a `apply` file in shipcat_cli to cover most of it, and more intelligently handle all errors from apply.

Then, rather than having multiple `upgrade_wrappers`, `reconcile_worker` functions that all basically do the same thing, we have one `apply` function that works regardless of whether or not you are coming from:

- shipcat apply
- shipcat cluster crd reconcile

Because it handles all the error cases. Note that this does not get rid of tiller yet. It just paves the way for us to do so by forcing versions into all CRDs at the apply/reconcile stage. Temporary, simplified forks of `helm::direct` functions now in `shipcat_cli::apply`.

Major changes:
- upgrade_wrapper deprecated
- most of shipcat_cli::helm module deprecated (generally only helpers useful now)
- `shipcat helm` commands mostly removed (will keep helm diff until kube 1.13 on eks)
- `shipcat diff` will replace the last `shipcat helm diff` (temporarily forked from helm::direct)
- `shipcat apply` is much smarter and deals with errors better
- Slack messaging has separated out what is CLI use and Event use (caused some issues)
- webhooks have added a new entrypoint from apply and propagated the new setup
- tons of functions in webhooks/audit marked as ready for deletion because of this
- UpgradeMode removed! Now only two bools: `force` and `wait`.
- UpgradeInfo is a new fork of UpgradeData (smaller, less special logic)
- UpgradeInfo deprecated (because it's in the helm module) - it actually handled shelling out to helm diff...
- Version inference has been standardised to be available on rolling environments.
- GET raftcat/version exposed (will start using crds after VERSION_URL decomission)
- cluster crd reconcile bypasses most of the logic to go straight into shipcat::apply (works fine)

Breaking changes:
- New reconcile mode -> new raftcat version required.
- apply no longer has --auto-rollback (change that one CI job that rely on it)


shipcat_cli/src/apply.rs
```diff
@@ -0,0 +1,317 @@
...
+    if let Err(e) = webhooks::ensure_requirements(&region) {
+        warn!("Could not ensure webhook requirements: {}", e);
+    }
+    let svc = mfbase.name.clone();
+    // Internal sanity
+    assert_eq!(mfbase.region, region.name);
+    assert_eq!(mfbase.namespace, region.namespace);
+
+    // A version can be explicit in two ways: passed via -t SOMEVER, or set in manifest
+    let explicit_version = override_version.or(mfbase.version.clone());
```
janslow: This behaviour already introduces some weird bugs. e.g., if you override a fixed version, it'll get rolled back on the next reconcile.

IMO, if a version is set in the manifest, it shouldn't be overridable:

```suggestion
    let explicit_version = mfbase.version.clone().or(override_version);
```

shipcat_cli/src/apply.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    // Figure out if the service is already installed
+    // This is currently needed to determine whether or not we run a diff.
+    // If diff could work this out (which in kube 1.13 we can), then:
+    // TODO: only do this inside hard case below (if no version given && environment.is_rolling())
```
janslow: Do we even need to `environment.is_rolling()`? It could just be called if `explicit_version.is_none()`, because that can only occur on rolling envs.

shipcat_cli/src/apply.rs
clux: Ah that's true. I'll put in a bail! for that.

shipcat_cli/src/apply.rs
clux: I can take it out if you feel this is not something shipcat should hardcode. I used it in show.rs to indicate an error case with `shipcat crd` (which cannot be piped in that case).

janslow: Sorry, I was thinking just about this section of the code, rather than `environment.is_rolling()` at large.

In particular, can we just wrap the (expensive) Helm call in an `if explicit_version.is_none()` statement

shipcat_cli/src/apply.rs
janslow: Thanks.

It's really confused me in the past, where I've accidentally hardcoded the version for dev and my deploys keep getting magically rolled back.

shipcat_cli/src/apply.rs
clux: We currently have to figure out if it was installed because we are required to know this before calling helm diff (it'll fail otherwise). We could call that speculatively, but that's the main expensive call we need to avoid :/

Have removed the `is_rolling` method on Environment though. That's more our convention I think.

PR #228, opened 2019-07-27
Fix compiler warnings

janslow:


shipcat_definitions/src/config.rs
```diff
@@ -443,7 +443,6 @@ impl Config {
     fn read_from(pwd: &PathBuf) -> Result<Config> {
         use std::fs::File;
         use std::io::prelude::*;
-        use semver::Version;
```
janslow: It's already defined at the top of the file

merge_derive/src/lib.rs
```diff
@@ -62,6 +62,7 @@ fn impl_merge_struct(fields: &Fields) -> TokenStream2 {
 
             for (i, field) in fields.iter().enumerate() {
                 if let Option::None = &field.ident {
+                    let i = syn::Index::from(i);
```
janslow: From https://github.com/rust-lang/rust/issues/60210

Issue #127, opened 2019-03-18
Support multiple Kong APIs per service


janslow:
Several services need to be exposed as multiple Kong APIs (e.g., they need different plugin configuration on different routes).

Currently, this requires the use of `extra_apis` on a region, which makes it harder to add new regions and track which service owns a region.

Instead, it would be useful if the `kong` property allowed multiple entries (Post #120, this could be done without complicating the `Manifest` struct).

For example, the following:

```yaml
# services/simple/shipcat.yml
kong:
  uris: '/simple'
  
# services/foo/shipcat.yml
kongs:
  ui: # API name suffix
    # enabled: true # implicit
    host: foo.example.com
    cookie_auth: true
  api:
    host: foo.example.com
    uris: '/api'
  old:
    host: old.example.com

# services/foo/dev-uk.yml
kongs:
  old:
    enabled: false
```

would map to the following kongfig (assuming `dev-uk`): 

```yaml
apis:
  - name: simple
    attributes: { uris: ['/simple'] }
  - name: foo-ui
    plugins: ... # cookie-auth
    attributes: { hosts: [foo.example.com] }
  - name: foo-api
    attributes: { hosts: [foo.example.com], uris: ['/api'] }
```

`kongs` is a map, as it allows the config to be overriden without having to redefine everything.

clux:
Yeah, that would be great. Extra apis in the config is horrible. They usually exist to facilitate cross-region communication, but afaikt are all owned by a service regardless, and there's no name consistency enforcement. It's been on our wishlist for a while.

clux:
Internally `Manifest::kongApis` is a vectorized thing now. Should this be closed?

janslow:
No, I haven't finished the `manifests` side logic.

It's a bit more complex, as we need to be able to have defaults defined at the global/region level which affect each API.

PR #230, opened 2019-07-27
Allow a CRD to have multiple Kong APIs

janslow:
Currently not exposed in manifest files, which I'll do in another PR (as that'll need some more complex logic to ensure the default Kong configuration is still inherited).

Part of #127

This is a breaking change, so will require a raftcat upgrade, but shouldn't crash old versions, because the `kong` field is being renamed to `kong_apis`, rather than changing the type.


janslow:
There's no change in the output of the following commands:

```
shipcat kong
shipcat get apistatus
shipcat statuscake
```

The output of `shipcat helm` and `shipcat crd` have changed for most services:

```diff
diff --git a/old-dev-uk.yaml b/new-dev-uk.yaml
index efca160b5..7e2ad0886 100644
--- a/old-dev-uk.yaml
+++ b/new-dev-uk.yaml
@@ -48,8 +48,8 @@ health:
   wait: 30
 httpPort: 80
 image: IMAGE_NAME
-kong:
-  add_headers:
+kongApis:
+- add_headers:
     Cache-Control: no-cache, no-store
```

shipcat_cli/src/get.rs
```diff
@@ -177,15 +177,15 @@ pub fn apistatus(conf: &Config, reg: &Region) -> Result<()> {
...
+        for k in mf.kongApis {
             let mut params = APIServiceParams {
                 uris: k.uris.unwrap_or("".into()),
                 hosts: k.hosts.join(","),
                 internal: k.internal,
                 publiclyAccessible: mf.publiclyAccessible,
                 websockets: false,
             };
-            if let Some(g) = mf.gate {
+            if let Some(g) = &mf.gate {
```
clux: ah, gate props. the mf.publiclyAccessible field is still dual defined :-/

shipcat_definitions/src/crds.rs
```diff
@@ -93,7 +93,7 @@ pub fn gen_all_crds() -> Vec<CrdSpec> {
                 name: "Kong".into(),
                 apcType: "string".into(),
                 description: "The URI where the service is available through kong".into(),
-                JSONPath: ".spec.kong.uris".into(),
+                JSONPath: ".spec.kong_apis[*].uris".into(),
```
clux: oh, how does this work? [*]?

shipcat_cli/src/kong.rs
```diff
@@ -70,14 +70,18 @@ pub fn generate_kong_output(conf: &Config, region: &Region) -> Result<KongOutput
...
+            if let Some(clash) = apis.insert(k.name.clone(), k) {
+                bail!("A Kong API named {:?} is already defined", clash.name);
+            }
         }
     }
 
     // Add general Kong region config
     for (name, api) in region.kong.extra_apis.clone() {
-        apis.insert(name, api);
+        if let Some(clash) = apis.insert(name, api) {
```
clux: huh, `insert` has this behaviour, interesting.

shipcat_definitions/src/manifest.rs
```diff
@@ -654,8 +654,8 @@ pub struct Manifest {
     ///   uris: /webapp
     ///   strip_uri: true
     /// ```
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub kong: Option<Kong>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub kongApis: Vec<Kong>,
```
clux: yeah, this is definitely breaking. crds need to be updated at the same time as raftcat and there'll likely be like 30s where raftcat is confused, but that's fine.

shipcat_filebacked/src/manifest.rs
```diff
@@ -184,11 +184,11 @@ impl ManifestSource {
             image: Some(self.build_image(&base.name)?),
 
             version: overrides.version.build(&())?,
-            kong: defaults.kong.build(&KongBuildParams {
+            kong_apis: defaults.kong.build(&KongBuildParams {
                 service: base.name.to_string(),
                 region: region.clone(),
                 hosts: overrides.hosts,
-            })?.unwrap_or(None),
+            })?.unwrap_or_default().map_or(Default::default(), |k| vec![k]),
```
clux: woah, dense line. feels like one `default` invocation here is redundant?

clux: presumably this is building "a default" and then vectorising it at the end?

shipcat_definitions/src/crds.rs
janslow: It's standard JSON path syntax, but unfortunately Istio will only return the first result: https://github.com/kubernetes/kubernetes/issues/67268#issuecomment-412282953

shipcat_cli/src/get.rs
janslow: Triple defined, there's also `kong.internal` (unless that's different).

shipcat_cli/src/kong.rs
janslow: Yeah, that way it's atomic, rather than getting and setting in two operations.

shipcat_filebacked/src/manifest.rs
janslow:
```suggestion
            })?.unwrap_or_default().map(|k| vec![k]).unwrap_or_default(),
```

is an alternative.

Unfortunately, `KongSource.build()` needs to return an option, because `kong: { host: null, uris: null }` is equivalent to `kong.enabled: false` for backwards compatibility.

https://github.com/Babylonpartners/shipcat/blob/master/shipcat_filebacked/src/kong.rs#L56

janslow: If we can migrate everyone to this new map based way of definining all of the APIs, we should be able to get rid of that check and just make it invalid to omit `host`/`uris`.

clux: Ah, double Option. Very well. I think your suggestion reads slightly better in that case, just because `map_or` is one of those you hardly ever see.

PR #231, opened 2019-07-27
Remove name field from KongSource struct

janslow:
It was never used, as the Kong API name currently comes from the service name.


janslow:
Blocked on https://github.com/Babylonpartners/manifests/pull/11953 (actually removing all usages of `kong.name` from the manifests)

PR #234, opened 2019-07-31
diff separation

clux:
- move diff helpers out of `helm::helpers` -> `diff`
- `helm` module now a single file
- remove duplicate `values` fns
- crd reconcile calls `apply` rather than via pointless indirection

best viewed in separate changesets


clux:
Slightly slowed down by https://github.com/Babylonpartners/shipcat/issues/235 as i wrote new tests :-(

Issue #191, opened 2019-05-15
Make Shipcat apply and CRD reconcile an atomic operation


janslow:
Currently, `shipcat cluster crd reconcile` is run first, then each service's Helm release is upgrade, if necessary.

This means, if the release upgrade fails, the CRD has already been updated, so Shipcat will think the service is up to date. Therefore it won't try to upgrade the release again on the next run, unless it's manually forced to do so.

Updating the CRD should be done atomically with the Helm upgrade (or at the very least, after the upgrade), to prevent this occurring.

clux:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status

```rust
pub struct ShipcatManifestStatus {
    pub conditions: Option<Vec<ManifestCondition>>,
    /// git version deployed from
    pub revision: String,
    // kong status?
    // canary status?
}

pub struct ManifestCondition {
    pub last_update_time: Option<Time>,
    pub message: String,
    pub reason: String,
    pub status: String,
    #[serde(rename = "type")
    pub type_: UpgradeState,
}

pub enum UpgradeState {
    // Complete, diff in reason, time taken to upgrade
    COMPLETE,
    // CRD applied with diff, diff in reason?
    STARTED,
    // main deployment failed to rollout in time (snapshot state in reason)
    TIMEOUT,
    // vault issues, chart issues, other uncaught issues (error in reason)
    ERROR,
}
```

janslow:
The status could also be useful for monitoring dashboards, which could tell if there is an error during deployments.

clux:
Have started fleshing out the .status object now. Was wondering how generic it was within kube so did a bunch of digging. The interested can read [this thread](https://github.com/clux/kube-rs/issues/43) on kube-rs.

A pr will appear here shortly regardless.

clux:
By the way the original bug has a workaround at the moment:

- `shipcat cluster crd reconcile` (fast reconcile, prone to problem)
- `SHIPCAT_MASS_RECONCILE=1 shipcat cluster crd reconcile` (slower, not prone to problem)

Can run the latter less frequently for now, but this is all going to get cleaned up once the `.status` objects are observeable.

PR #238, opened 2019-08-01
Start writing shipcatmanifest .status on apply

clux:
Fixes https://github.com/Babylonpartners/shipcat/issues/191 when this is complete.

~~WIP ATM. It mocks core stuff heavily. But I've raised a draft PR so people can comment on field names and stuff~~. Chose not to use the normal vectorised condition types because [they are really hard to use well](https://github.com/clux/kube-rs/issues/43).

This is now usable.  Got rid of JSON patch (it was horrible), and you can actually null things out with MERGE patch, provided you don't provide the serde skip_serializing_if annotation.


shipcat_cli/src/apply.rs
```diff
@@ -208,6 +228,202 @@ pub fn apply(svc: &str,
...
+    let mf = mfcrd.clone().complete(&region).map_err(|e| {
+        // Fire failed events if secrets fail to resolve
+        webhooks::apply_event(UpgradeState::Failed, &ui, &region, &conf);
+        // TODO: set up retry based on previous .status states here..
+        e
+    })?;
+
+    // Create values file
+    let hfile = format!("{}.helm.gen.yml", svc);
+    helm::values(&mf, &hfile)?;
```
gleaver: Missing error handling

shipcat_cli/src/apply.rs
```diff
@@ -208,6 +228,202 @@ pub fn apply(svc: &str,
...
+            let fallback = o.spec.version.clone().expect("manifest.version set");
+
+            // A version was supplied, and it differs from active CRD
+            if explicit_version != o.spec.version {
+                reason = Some(UpgradeReason::VersionChange);
+            }
+            (explicit_version.unwrap_or(fallback), Some(o))
+         },
+    };
+    let exists = crd.is_some(); // no crd => nothing installed
```
gleaver: Be more explicit about old CRD vs new CRD variable naming.

shipcat_cli/src/apply.rs
```diff
@@ -179,21 +203,17 @@ pub fn apply(svc: &str,
...
-                            webhooks::apply_event(UpgradeState::Failed, &ui, &region, &conf);
-                            return Err(ErrorKind::UpgradeTimeout(mf.name.clone(), mf.estimate_wait_time()).into());
-                        }
+                match kubectl::await_rollout_status(&mf) {
+                    Ok(true) => {
+                        info!("successfully rolled out {}", &ui.name);
+                        webhooks::apply_event(UpgradeState::Completed, &ui, &region, &conf);
+                    },
+                    Ok(false) => {
+                        let _ = kubectl::debug_rollout_status(&mf);
```
janslow: What does this do? Is it just printing out a log message?

Also, why are we discarding the result?

shipcat_cli/src/apply.rs
```diff
@@ -208,6 +228,202 @@ pub fn apply(svc: &str,
...
+    // - if we need to interact with secret-manager TODO: do
+    let s = status::Status::new(&mfbase)?;
+
+    // Next large batch is working out the reason for the upgrade (if any)
+    let mut reason = None;
+    // Fetch the full shipcatmanifest crd
+    // NB: if kube api or early stuff here fails, then we are in a bad state.
+    // Before we have applied any CRDs it's safe to bail. Next run can retry.
+    let (actual_version, crd) = match s.get() {
+        Err(_e) => {
```
janslow: Can we log the error as a warning? I can imagine this suppressing a more important error and making debugging a pain (e.g., an RBAC misconfiguration).

shipcat_cli/src/apply.rs
janslow: It'll propagate the error, due to the `?`. Or do you mean wrapping the error with a more useful message?

shipcat_cli/src/apply.rs
```diff
@@ -208,6 +228,202 @@ pub fn apply(svc: &str,
...
+    match upgrade_helm(&mf, &hfile) {
+        Err(e) => {
+            let reason = e.description().to_string();
+            let _ = s.update_apply_false(ureason.to_string(), "ApplyFailure", reason);
+            error!("{} from {}", e, ui.name);
+            webhooks::apply_event(UpgradeState::Failed, &ui, &region, &conf);
+            return Err(e);
+        },
+        Ok(_) => {
+            let _ = s.update_apply_true(ureason.to_string());
```
janslow: Why are we discarding the return value here?

shipcat_cli/src/apply.rs
```diff
@@ -208,6 +228,202 @@ pub fn apply(svc: &str,
...
+    }
+
+    // We cannot be here without a reason now, although you have to convince yourself.
+    let ureason = reason.expect("cannot apply without a reason");
+
+    webhooks::apply_event(UpgradeState::Pending, &ui, &region, &conf);
+    match upgrade_helm(&mf, &hfile) {
+        Err(e) => {
+            let reason = e.description().to_string();
+            let _ = s.update_apply_false(ureason.to_string(), "ApplyFailure", reason);
```
janslow: Why are we discarding the return value here?

shipcat_cli/src/apply.rs
```diff
@@ -208,6 +228,202 @@ pub fn apply(svc: &str,
...
+        },
+        Ok(_) => {
+            let _ = s.update_apply_true(ureason.to_string());
+            if !wait {
+                info!("successfully applied {} (without waiting)", ui.name);
+            }
+            else {
+                match kubectl::await_rollout_status(&mf) {
+                    Ok(true) => {
+                        let _ = s.update_rollout_true();
```
janslow: Why are we discarding the return value here?

shipcat_cli/src/status.rs
```diff
@@ -0,0 +1,362 @@
...
+    // TODO: vault secret hash hash
+
+    // MAYBE: kong status?
+    // MAYBE: canary status?
+}
+
+#[derive(Deserialize, Serialize, Clone, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct Conditions {
+    /// Apply status
```
gleaver: Add comments about specifically what apply/rollout exactly mean.

shipcat_cli/src/status.rs
```diff
@@ -0,0 +1,362 @@
...
+};
+
+/// Status object for shipcatmanifests crd
+///
+/// All fields optional, but we try to ensure all fields exist.
+#[derive(Deserialize, Serialize, Clone, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct ManifestStatus {
+    /// Detailed individual conditions, emitted as they happen during apply
+    pub conditions: Option<Conditions>,
```
janslow: Can this not be defaulted?

```suggestion
	#[serde(default)]
    pub conditions: Conditions,
```

Because

```rs
None == Some(Conditions { apply: None, rollout: None }) == Some(Conditions::default())
```

Therefore there's no need to unwrap the option elsewhere

shipcat_cli/src/status.rs
```diff
@@ -0,0 +1,362 @@
...
+
+    /// Best effort most relevant reason for why an apply failed
+    ///
+    /// This is nulled out when rollout condition is set to true
+    #[serde(default)]
+    reason: Option<String>,
+
+    /// Best effort reason for why an apply was triggered
+    #[serde(default)]
+    upgrad_reason: Option<String>,
```
gleaver: UPGRAD

shipcat_cli/src/apply.rs
clux: I'm making a generated condition for these types of errors.

shipcat_cli/src/apply.rs
clux: It prints to stdout speculatively. Don't want to propagate any errors about it.
Was considering capturing some of the information and putting it on the .status object, but that's a different discussion.

shipcat_cli/src/apply.rs
clux: It's already logged from s.get() atm, and i'm having some issues with failure + error_chain incompatibility. Ideally i'd just print this` _e` here, and not log from `get()`, but I need to clean up error handling properly across the CLI, and that's a bigger problem.

shipcat_cli/src/apply.rs
clux: I'll catch this error here. Leftover carefulness from when i didn't feature flag rollout this.

shipcat_cli/src/apply.rs
clux: Ditto. Will propagate.

shipcat_cli/src/status.rs
clux: Just because it has a default doesn't mean it's sensible to treat a missing value as such. A default condition would imply a concrete timestamp here (which I don't want).

shipcat_cli/src/apply.rs
clux: ditto

shipcat_cli/src/apply.rs
clux: ditto

shipcat_cli/src/apply.rs
clux: Doing this last so i can absorb the error. Shipcat failing is more important.

shipcat_cli/src/status.rs
janslow: A default `Condition` has a timestamp, but a default `Conditions` has `None` for each `Condition`

clux: Ah. true!

shipcat_cli/src/apply.rs
```diff
@@ -353,14 +357,15 @@ fn apply_with_crd(svc: &str,
 
     // We cannot be here without a reason now, although you have to convince yourself.
     let ureason = reason.expect("cannot apply without a reason");
+    s.update_generate_true()?;
 
     webhooks::apply_event(UpgradeState::Pending, &ui, &region, &conf);
```
gleaver: double pending

PR #240, opened 2019-08-04
Remove legacy OAuth2 properties from Kong config

janslow:
`kong.oauth2_anonymous` and `kong.oauth2_extension_plugin` are no longer needed, as they've been superseded by the `authorization` section.

All usages of the properties have now been removed, so we can safely delete them.

In addition, we can now remove the `oauth2-extension` plugin from Kong.


janslow:
I've diffed the output of the `kong` command in each environment, the only change is the removal of the following from each API:

```diff
@@ -23550,8 +23160,6 @@ apis:
     name: tcp-log
   - ensure: removed
     name: oauth2
-  - ensure: removed
-    name: oauth2-extension
   - attributes:
       config:
         anonymous_username: anonymous
```

PR #245, opened 2019-08-05
[IDM-302] Remove `Kong#cookie_auth*` fields

janslow:
Superseded by `authorization.allow_cookies`.

The original property allowed cookie auth to be enabled without CSRF protection, which is bad.


janslow:
* [ ] Blocked on [removal of last usage](https://github.com/Babylonpartners/manifests/pull/12290)

PR #248, opened 2019-08-06
Better shipcat status

clux:
shpcat-status


clux:
mostly stole kpods:

```
kpods() {
  # https://github.com/kubernetes/kubernetes/issues/71612
  local -r app="$1"
  paste <(kubectl get pods -lapp="${app}" --sort-by=.metadata.creationTimestamp) \
        <(kubectl get pods -lapp="${app}" -o=custom-columns=VERSION:"{spec.containers[*].env[?(@.name=='SERVICE_VERSION')].value}",ENV:metadata.namespace --sort-by=.metadata.creationTimestamp)
}
```

and made it a bit smarter... it avoids the race condition in that command, but it doesn't count ready stuff correct. Extra columns isn't a thing atm.

PR #257, opened 2019-08-10
make cargo clippy pass with 3 disabled lints

clux:
`make clippy` encapsulates the allow calls. They are as follows:

- [clippy::or_fun_call](https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call) - makes lines longer and harder to understand
- [clippy::redundant_pattern_matching](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pattern_matching) - annoyingly forces you to swap between `if let` and some function suffix on `Option` or `Result` - even if you discard the let binding
- [clippy::redundant_field_names](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_field_names) - if there's consistency in the function, that should trump the tiny gain you get from this..

Additionally, a few allow instructions added to the code where it makes sense. Commenting on them specifically.


shipcat_cli/src/graph.rs
```diff
@@ -135,6 +135,7 @@ pub fn full(dot: bool, conf: &Config, reg: &Region) -> Result<CatGraph> {
                 trace!("Found dependency new in graph: {}", dep.name);
                 let depmf = shipcat_filebacked::load_manifest(&dep.name, conf, reg)?;
                 let depnode = ManifestNode::new(&depmf);
+                #[allow(clippy::let_and_return)]
                 let depidx = graph.add_node(depnode);
                 depidx
```
clux: not really intuitive that add_node would return an index here, and we have to scan far up to see it..

shipcat_cli/src/main.rs
```diff
@@ -410,12 +410,13 @@ fn resolve_config(args: &ArgMatches, ct: ConfigType) -> Result<(Config, Region)>
...
 }
 
-fn void<T>(_x: T) { () } // helper so that dispatch_commands can return Result<()>
+fn void<T>(_x: T) {} // helper so that dispatch_commands can return Result<()>
 
 /// Dispatch clap arguments to shipcat handlers
 ///
 /// A boring and somewhat error-prone "if-x-then-fnx dance". We are relying on types
 /// in the dispatched functions to catch the majority of errors herein.
+#[allow(clippy::cognitive_complexity)]
```
clux: big fn, but it's all doing the same type of thing. it's never broken more than the subcommand you've touched.

shipcat_definitions/src/structs/kongfig.rs
```diff
@@ -320,6 +320,7 @@ impl Default for CorrelationIdPluginConfig {
     }
 }
 
+#[allow(clippy::large_enum_variant)]
```
clux: wanted us to Box things. this probably doesn't matter a lot though..

shipcat_filebacked/src/container/mod.rs
```diff
@@ -1,9 +1,11 @@
+#[allow(clippy::module_inception)] // not exporting the module..
 mod container;
+pub use container::ContainerBuildParams;
```
clux: container mod inside a container mod, kind of funny error, but doesn't matter.

PR #265, opened 2019-08-18
Deprecating old apply

clux:
Also provide -t for template


shipcat_cli/src/apply.rs
```diff
@@ -91,16 +91,18 @@ pub fn apply(svc: &str,
 }
 
 
-/// Current legacy version
+/// Legacy version
 ///
 /// Reads CRDs, but doesn't write to them. Upgrades through tiller.
+#[deprecated(since = "0.132.2", note = "CrdVersioned users should move on to CrdStatus")]
```
janslow: Is there a reason for not just deleting it?

clux: I'll pm you.

PR #266, opened 2019-08-18
make a parallel shipcat validate (verify)

clux:
this will eventually deprecate validate.
can run verify on all regions in about 3s locally.


shipcat_cli/src/validate.rs
```diff
@@ -1,5 +1,63 @@
...
+use std::sync::mpsc::channel;
+
+/// Validate all manifests in a service directory for a region
+///
+/// This is meant to replace `shipcat validate ..all_services`
+pub fn regional_manifests(conf: &Config, reg: &Region) -> Result<()> {
+    let available = shipcat_filebacked::available(conf, &reg)?;
+
+    let len = available.len();
+    let pool = ThreadPool::new(16); // 16 manifests per worker
```
gleaver: stick magic number in a named constant?

shipcat_cli/src/validate.rs
```diff
@@ -1,5 +1,63 @@
...
+
+/// Validate all manifests in a service directory for ALL regions
+///
+/// This is meant to replace a for loop over shipcat list-regions
+pub fn all_manifests() -> Result<()> {
+    use crate::ConfigType;
+    let regions = Config::read()?.list_regions();
+
+    let len = regions.len();
+    let pool = ThreadPool::new(len); // all regions in parallel
```
gleaver: Is this additional threadpool actually doing anything useful?

clux: there's two disjoint pools atm, one pool makes 16 more.

clux: moved to rayon.

PR #267, opened 2019-08-21
Fixes for speedups

clux:
- missing good xargs (no -P unless you have findutils)
- fixing circle caching to do what ewok does
- allowing language to be put on .metadata (deprecating it) - will move it between releases


shipcat_definitions/src/structs/metadata.rs
```diff
@@ -12,6 +12,9 @@ pub struct Contact {
     pub name: String,
     /// Slack handle
     pub slack: String,
+    /// Language the service is written in
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub language: Option<String>,
```
gleaver: Is this worth enumming to make sure it can only be one of a few values? There's some useful stuff we could predicate off this if it's a well-known thing.

gleaver: ...shouldn't this be in metadata?

janslow: I think that'd mean that adding a new enum value would break the Rust API (e.g., Raftcat would fail to parse the new value).

clux: i can make an enum for it

.circleci/config.yml
```diff
@@ -42,25 +42,19 @@ jobs:
     docker:
       - image: clux/muslrust:stable
     working_directory: /volume
-    parallelism: 2
     steps:
       - checkout
+      - run: grep "\[metadata\]" -A 9999 Cargo.lock > deps_checksum
```
gleaver: disgusting

.circleci/config.yml
```diff
@@ -118,28 +106,18 @@ jobs:
...
-    parallelism: 2
     steps:
       - checkout
-      - run:
-          name: Install rust
-          command: |
-            curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable
-      - restore_cache:
-          key: cargo.registry-mac-{{ .Environment.CACHE_VERSION }}
+      - run: curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable
```
gleaver: Is there not a standard docker image with an up to date toolchain that doesn't require a curlbash?

clux: not in macos land!

.circleci/config.yml
clux: i only regret going 999 above what clearly is the right number

.circleci/config.yml
```diff
@@ -72,33 +67,26 @@ jobs:
...
     docker:
-      - image: clux/muslrust:stable
-    working_directory: /volume
+      - image: circleci/rust:latest
     resource_class: xlarge
     environment:
       - RUST_BACKTRACE: "full"
     steps:
       - checkout
+      - run: grep "\[metadata\]" -A 9000 Cargo.lock > deps_checksum
```
janslow: You could extract the caching logic into a custom command, to simplify the jobs themselves

e.g., https://github.com/Babylonpartners/identity-cli/blob/master/.circleci/config.yml#L201

clux: true. the paths are all docker image dependent though :-(

janslow: You can use parameters: https://github.com/Babylonpartners/identity-cli/blob/master/.circleci/config.yml#L234

janslow: Actually, https://github.com/Babylonpartners/manifests/blob/master/.circleci/config.yml#L162 is probably a better example

clux: I'll look at this another time, already yaml'ed out today.

PR #268, opened 2019-08-24
Products in manifest syntax

clux:
To allow categorisation of services.

**Please note that the raftcat changes are mostly due to rustfmt (which is gradually being added)**

Some small tweaks to languages necessary while this was attempted to be ported in https://github.com/Babylonpartners/manifests/pull/13064


shipcat_definitions/src/structs/metadata.rs
```diff
@@ -88,6 +89,13 @@ pub enum Language {
     TypeScript,
     Kotlin,
     Swift,
+    Php,
+    Elixir,
+    Clojure,
+    Haskell,
+    C,
+    Cpp,
+    Bash,
```
janslow: So a bash service isn't weird? 😄

clux: oh for sure, i just put top 20 languages in there :slightly_smiling_face:

PR #274, opened 2019-09-03
reconcile mode for kube 1.12

clux:
Fixes bugs where people can set conflicting default urls in kong.

`reconcilationMode` is now the properly rolled out `CrdStatus`. Added a changelog for this stuff since we're now making some breaking changes. Good to document them.


CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+0.126.0 / 2019-09-03
+==================
+  * `reconciliationMode` default is now `CrdStatus` rather than `CrdVersioned`
```
janslow: If you make sure that the squashed commit messages are useful, you can just get `ghr` to use those for release notes https://github.com/Babylonpartners/identity-cli/releases/tag/1.0.1

shipcat_cli/src/validate.rs
```diff
@@ -10,22 +10,34 @@ use rayon::prelude::*;
...
+        .partition_map(Either::from);
 
     if !errs.is_empty() {
         for e in &errs {
             error!("{}", e);
         }
         bail!("Invalid shipcat data in {} files", errs.len());
     }
+    // cross reference uniqueness values here
+    let mut uris = oks.iter()
```
janslow: You need to include the hosts in there as well.

CHANGELOG.md
clux: this is more for highlighting breaking changes

shipcat_cli/src/validate.rs
clux: as in, the comma separated list of kong hosts should also be validated?

clux: added validation of hosts, but it ignores host checks

janslow: Sorry, it's the tuple of `(host, path)` which needs to be unique.

E.g., the `workflow-engine` has two APIs on it's own host (one has a path, the other doesn't).

clux: that's a lot more complicated to check for. hosts is a 2 level deep thing and uris is a 1 level deep thing. i'm not sure how i am supposed to zip it to get the check i want. should i associate a k.uris entry with every k.hosts entry and expand?

janslow: Yep.

As a first step, you could just ignore APIs with a specific host (i.e., only look at APIs for the service domain), which is what the Gate config generator does atm (https://github.com/Babylonpartners/manifests/blob/master/scripts/gate-config-generator.py#L58).

The APIs with `hosts` defined don't need the validation as much, because:
* There's less of them
* A host is generally owned by a single team, so they're more likely to be aware of the clash

clux: Ok, if the gate config is already doing some validation on this i am happy to leave it for now.

doc/extending.md
```diff
@@ -43,7 +43,7 @@ impl Dependency {
         }
         // self.api must parse as an integer
         assert!(self.api.is_some(), "api version set by implicits");
-        if let Some(ref apiv) = self.api {
+        if let Some(apiv) = &self.api {
```
janslow: What's the difference between these?

clux: abbreviation really, ergonomics thing in some recent stable

janslow: Ok, I was wondering if there was a difference, as I've not used `ref`

PR #282, opened 2019-09-17
allow noop version overrides in shipcat apply

clux:


akoss:
Looks good, makes sense

PR #284, opened 2019-09-21
[IDP-290] kongfig changes for Kong json-cookie-to-headers refresh support

akoss:
Babylon's Kong json-cookie-to-headers plugin now supports refreshing expired access tokens automatically. This PR adds the related `kongfig` fields. 

Note: to make sure that we are able to rollback to an older version of Kong, all new fields are optional and removed from the kongfig output when unset. (Extra parameters are rejected by `kongfig`.) Legacy fields both in JsonCookiesCsrfPluginConfig and JsonCookiesToHeadersPluginConfig are kept for the same reason.


shipcat_filebacked/src/authorization.rs
```diff
@@ -26,12 +33,24 @@ impl Build<Authorization, ()> for AuthorizationSource {
...
+            bail!("enable_cookie_refresh requires refresh_auth_service and refresh_body_refresh_token_key");
+        }
+
         Ok(Authorization {
             allowed_audiences,
             allow_anonymous,
             allow_invalid_tokens,
             required_scopes: self.required_scopes.unwrap_or_default(),
             allow_cookies: self.allow_cookies.unwrap_or_default(),
+            enable_cookie_refresh: enable_cookie_refresh,
```
clux: you can put this at the top as just `enable_cookie_refresh,` because key name == value name

shipcat_definitions/src/structs/kongfig.rs
```diff
@@ -413,7 +433,22 @@ pub fn kongfig_apis(from: BTreeMap<String, Kong>, config: KongConfig, region: &R
...
+                        
+                        auth_service: a.refresh_auth_service,
+                        body_refresh_token_key: a.refresh_body_refresh_token_key,
+                        cookie_max_age_sec: a.refresh_max_age_sec,
+                        enable_refresh_expired_access_tokens: Some(a.enable_cookie_refresh),
+                        http_timeout_msec: a.refresh_http_timeout_msec,
+                        renew_before_expiry_sec: a.refresh_renew_before_expiry_sec,
+                    })));
+                } else {
+                    plugins.push(ApiPlugin::JsonCookiesToHeaders(PluginBase::default()));
```
janslow:
```suggestion
                    plugins.push(ApiPlugin::JsonCookiesToHeaders(PluginBase::removed()));
```

PR #286, opened 2019-09-23
[DEVP-979] Move all git logic into a module, and make diffing command…

gleaver:
…s operate against merge-base origin/master

Some tidying up to do on this (versions/docstrings), but basically complete. Only tangible change is to make git operations do `git diff $(git merge-base origin/master HEAD)` rather than `git diff master`. Tidied all the git commands into a single module to remove some duplication and encapsulate commands.


shipcat_cli/src/validate.rs
```diff
@@ -165,25 +166,14 @@ pub fn config(conf: Config) -> Result<()> {
...
-        bail!("Subprocess failure from git: {}", s.status.code().unwrap_or(1001))
-    }
-    let out : String = String::from_utf8_lossy(&s.stdout).into();
-    let err : String = String::from_utf8_lossy(&s.stderr).into();
-    if !err.is_empty() {
-        warn!("{} stderr: {}", args.join(" "), err);
-    }
-    debug!("{}", out);
+    let merge_base = git::merge_base()?;
+    let diff_output = git::diff_filenames(merge_base.as_str())?;
```
clux: does `&merge_base` not work here?

PR #287, opened 2019-09-25
[IDP-290] fixing recursin in JsonCookiesToHeadersPluginConfig definition (broke 0.128.0)

akoss:


shipcat_definitions/src/structs/kongfig.rs
```diff
@@ -301,7 +301,13 @@ impl Default for JsonCookiesToHeadersPluginConfig {
             // Deprecated fields - to be removed after Kong update
             field_name: Some("kong_token".into()),
             cookie_name: Some("autologin_token".into()),
-            ..Default::default()
```
clux: Ooh, right. Yeah, you'd probably need `JsonCookiesToHeadersPluginConfig::default` here. I assumed you had tested the output of `shipcat kong` before merging.

akoss: Ahh yeah I tested thoroughly but not after adding this. Seemed harmless enough and automated tests were passing. Thanks 👍

clux: Yeh, understandable.

PR #289, opened 2019-09-26
[DEVP-979] Add --with-region option to diff to allow diffing of untem…

gleaver:
…plated values between two regions.

New feature: `shipcat diff chatscript --region prod-uk --with-region preprod-uk`. Generates the base manifest for both, and diffs the values (for output with danger).

Also generalised the diff function to allow the filenames to be specified, so the diff will read:
```diff
--- diagnostic-engine.preprod-uk.shipcat.gen.yml        2019-09-26 10:53:44.000000000 +0100                                                                                                                                                                                                   
+++ diagnostic-engine.prod-uk.shipcat.gen.yml   2019-09-26 10:53:44.000000000 +0100                                                                                                                                                                                                           
@@ -10,7 +10,7 @@                                                                                                                                                                                                                                                                             
   description: ~                                                                                                                                                                                                                                                                             
 chart: base                                                                                                                                                                                                                                                                                  
 image: quay.io/babylonhealth/diagnostic-engine                                                                                                                                                                                                                                               
-version: 4.55.0                                                                                                                                                                                                                                                                              
+version: 4.54.0 
```


shipcat_cli/src/main.rs
```diff
@@ -307,6 +307,14 @@ fn build_cli() -> App<'static, 'static> {
                 .long("git")
                 .global(true)
                 .help("Comparing with master using a temporary git stash and git checkout"))
+              .arg(Arg::with_name("with-region")
+                .long("with-region")
+                .global(true)
+                .takes_value(true)
+                .conflicts_with("git")
+                .conflicts_with("crd")
```
janslow: Surely this doesn't conflict with `--crd`, it implies it.

gleaver: My reasoning was that like the `--git` arg, this is a different behaviour of diff, which is offline. I'm happy either way if you think it makes sense to remove it.

janslow: I'm not too bothered either way.

gleaver:
Immediate use case: https://github.com/Babylonpartners/manifests/pull/14396

PR #291, opened 2019-09-27
make region.kong actually optional

clux:
sure, you can fill the region struct with duds..
but the real solution is using an optional.

**Change:**
- `region.kong` is now an `Option<KongConfig>` rather than a `KongConfig`
- propagate.

Some explicit bails when relevant cmds use it and there is no kong in the region. which is fine. explicit error is better than accidental use of dud values.

yak-shave.


shipcat_filebacked/src/kong.rs
```diff
@@ -40,42 +40,46 @@ impl Build<Option<Kong>, KongBuildParams> for KongSource {
     /// Build a Kong from a KongSource, validating and mutating properties.
     fn build(self, params: &KongBuildParams) -> Result<Option<Kong>> {
         let KongBuildParams { region, service } = params;
-        let hosts = self.build_hosts(&region.kong.base_url)?;
+        if let Some(k) = &region.kong {
```
janslow: It should be an error if you try to define a Kong API in a region which doesn't have a Kong.

shipcat_cli/src/kong.rs
```diff
@@ -64,26 +64,29 @@ impl KongCrdOutput {
...
+pub fn generate_kong_output(conf: &Config, region: &Region) -> Result<Option<KongOutput>> {
     let mut apis = BTreeMap::new();
-
-    // Generate list of APIs to feed to Kong
-    for mf in shipcat_filebacked::available(conf, region)? {
-        debug!("Scanning service {:?}", mf);
-        for k in mf.kong_apis {
-            if let Some(clash) = apis.insert(k.name.clone(), k) {
-                bail!("A Kong API named {:?} is already defined", clash.name);
+    if let Some(kong) = &region.kong {
```
janslow: I feel this should also bail if there's no Kong.

e.g., `shipcat kong -r dev-ops` should fail

shipcat_filebacked/src/kong.rs
clux: Yeah, I agree. Not sure where it's best to do that though. If I do it in this function you can't create any manifest without having kong parameters set on the Region.

shipcat_cli/src/kong.rs
clux: This is technically the case from the entrypoint, but have lifted the error up rather than return an Optional from the helper function.

janslow:
```
let kong = match &region.kong {
  Some(k) => k,
  None => bail!("kong not available in {}", region.name),
};
```

To unwrap the if statement (keeps the complexity lower).

Or `&region.kong.ok_or(Err(format!("kong not available in {}", region.name)))?`

shipcat_definitions/src/structs/security.rs
```diff
@@ -174,11 +174,11 @@ impl DataHandling {
                 let enc = f.encrypted.unwrap(); // filled by implicits
                 // can't block on this yet - so just warn a lot
                 if f.name.is_spii() && !enc {
```
janslow: Is there any point in having these checks?

shipcat_filebacked/src/kong.rs
janslow:
```rs
if self.hosts.is_empty() && self.uris.is_none() {
  return Ok(None);
}

let kong = match &region.kong {
  Some(k) => k,
  None => bail!("kong not available in {}", region.name),
};

let hosts = self.build_hosts(kong.base_url)?;
```

This should enforce it without requiring any changes.

janslow: (it's just pulling the condition up a bit)

shipcat_definitions/src/structs/security.rs
clux: not atm no. left it in there due to the original requirements, but as it stands the dataHandling stuff is all unused in terms of gdpr.

shipcat_cli/src/kong.rs
clux: You would need an explicit error kind for this pattern though, which results in a much larger line to scan. I'd prefer to just have the if let Some() and a much simpler `bail!` in the else case tbh

shipcat_definitions/src/structs/security.rs
clux: have marked is as experimental everyhwere on EG.

shipcat_filebacked/src/kong.rs
clux: hosts is grabbed from region.kong though so you need to do the match first

clux:
Allows https://github.com/Babylonpartners/manifests/pull/14622/files

PR #293, opened 2019-10-01
DEVP-995 extra deps for python tests and linter

bh-anton-k:


Dockerfile.circleci
```diff
@@ -33,13 +33,17 @@ RUN set -x && \
...
+        "boto3>=1.9.239" \
+        "prettytable>=0.7.2" \
         "colorama>=0.3.7,<1.0.0" \
         "jinja2>=2.10,<3.0" \
         "jira>=2.0.0,<3.0.0" \
         "mock>=2.0.0,<3.0.0" \
         "packaging>=19.0,<20.0" \
         "PyGithub>=1.43.4,<2.0.0" \
         "pytest>=5.0.0,<6.0.0" \
+        "pytest-cov>=2.7.1" \
```
gleaver: It's good practice to prevent the upgrade to a major version, as that might break things: `pytest-cov>=2.7.1,<3.0.0`

PR #301, opened 2019-10-07
add hub and spoke struct

RenatoSerra22:


clux:
Going to relegate this to @akoss and/or @janslow .

Issue #300, opened 2019-10-07
shipcat_filebacked should not construct if unsupported region


clux:
Currently `shipcat validate` fails when `-r` points to an undeployed region, but `shipcat values` and `shipcat template` will happily spit out an extrapolated guess.

Filebacked should fail in this case and produce an error like `validate`:

```sh
$ shipcat validate raftcat -r dev-internal
shipcat: validate error: Unsupported region dev-internal for service raftcat
```

clux:
Think we actually fixed this in the last error handling issue and forgot to close.


```
$ shipcat validate raftcat -r dev-internal
shipcat: validate error: Unsupported region dev-internal for service raftcat
```

PR #302, opened 2019-10-07
[DEVP-1023] Update diff --git and --with-region to show full diff whe…

gleaver:
…n a service added or removed. Also bail on loading a manifest from a region where it does not exist.

Catch an edge case where if services are added or removed we fail to diff them correctly. This happens because we consider a filebacked manifest to be valid even if the region it is loaded for is not used.

This change bails if attempting to load a manifest for an unused region, and amends the CRD diff functionality to consider the lack of a region to be an empty string.

Feedback appreciated on ways in which this might break things/not behave as expected.


gleaver:
Altered implementation to only verify the region is correct in the diff functionality, as verifying during load seemed too low-level a place to put the bail.

gleaver:
Related manifests change: https://github.com/Babylonpartners/manifests/pull/14982

shipcat_cli/src/diff.rs
```diff
@@ -8,15 +8,30 @@ use super::{Config, Region, Manifest, Result};
...
+/// otherwise YAML serialise the content. For diff purposes, the content
+/// of a manifest not in a region is a blank, rather than being invalid.
+fn as_yaml(svc: &str, conf: &Config, region: &Region) -> Result<String> {
+    let mf = shipcat_filebacked::load_manifest(&svc, conf, region)?;
+    if let Ok(m) = mf.verify_region() {
+        let yaml = serde_yaml::to_string(&m)?;
+        Ok(yaml)
+    } else {
+        Ok("".to_string())
+    }
```
clux: wait `load_manifest` works in a region that it's not included in?
i guess that's part of the #300 bug?

gleaver: That's the bug, yeah.

PR #304, opened 2019-10-08
Bumping version to 0.132.0

akoss:


clux:
Hang on a second, I'm making a tweak to rbac. Will shoe-horn that in.

PR #306, opened 2019-10-08
better example docs

clux:
no need to talk about tiller anymore


clux:
Thanks. Added some clarifying instructions.

PR #307, opened 2019-10-09
[DIPH-138] use which to ensure the helm executable exists on system

bclewis1:


bclewis1:
It took me ages to figure out why `shipcat template` wasn't running on my machine before I realised that it depends on `helm`. I thought that adding a little warning might be helpful, although I couldn't find anything in the standard library to do this.

bclewis1:
without the fix:
```
➜  shipcat git:(master) RUST_LOG=debug shipcat template -ds -r staging-uk core-auth-2
shipcat (line 412): template error: No such file or directory (os error 2)
```

with the fix:
```
➜  shipcat git:(DIPH-138-improve-error-messaging) RUST_LOG=debug shipcat template -ds -r staging-uk core-auth-2
shipcat (line 412): template error: helm executable not found!
```

PR #308, opened 2019-10-09
DEVP-1024 monitoring model

bh-anton-k:


shipcat_definitions/src/structs/monitoring.rs
```diff
@@ -0,0 +1,172 @@
...
+        }
+        Ok(())
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct Sentry {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub enabled: Option<bool>,
```
janslow: `monitoring.sentry == None` is equivalent to `monitoring.sentry = Sentry { enabled: false }`, so this property should be removed.

Otherwise, consumers have to deal with both cases (this can be handled in `shipcat_filebacked`)

shipcat_definitions/src/structs/monitoring.rs
```diff
@@ -0,0 +1,172 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub envName: Option<String>,
```
janslow: `envName` seems a bit ambiguous (environment is a very overloaded term).

`dsnEnvVar` or similar would be more obvious

shipcat_definitions/src/structs/monitoring.rs
```diff
@@ -0,0 +1,172 @@
...
+        Ok(())
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct NewRelicAlert {
+    pub name: String,
+    #[serde(skip_serializing_if = "Option::is_none", rename = "type")]
+    pub alertType: Option<NewRelicAlertType>,
```
janslow: Is this actually optional?

shipcat_definitions/src/structs/monitoring.rs
janslow: Also, the env vars should be snake_case (https://doc.rust-lang.org/1.0.0/style/style/naming/README.html)

shipcat_definitions/src/structs/monitoring.rs
```diff
@@ -0,0 +1,172 @@
...
+#[serde(deny_unknown_fields)]
+pub struct Sentry {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub enabled: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub envName: Option<String>,
+}
+
+impl Sentry {
+    pub fn verify(&self) -> Result<()> {
```
janslow: The verification should be run in `shipcat_filebacked` (e.g., https://github.com/Babylonpartners/shipcat/blob/master/shipcat_filebacked/src/container/source.rs#L17)

shipcat_definitions/src/manifest.rs
```diff
@@ -683,6 +683,33 @@ pub struct Manifest {
...
+    ///          threshold: "0.5"
+    ///          time_function: any
+    ///        - duration: "100"
+    ///          operator: below
+    ///          priority: critical
+    ///          threshold: "0.2"
+    ///          time_function: any
+    /// ```
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub monitoring: Option<Monitoring>,
```
janslow: `monitoring: None` is equivalent to `monitoring: Some(Monitoring::default())`, so there's no need to wrap this in an `Option`

bh-anton-k: thanks, don't take it too seriously, yet, I aim to test-cover it and tighten the definitions accordingly

shipcat_definitions/src/structs/monitoring.rs
bh-anton-k: well there're some speculations going on - we're now supporting only one type and some other would need extra fields supplied, so I defined all known valid values for the field though for current implementation we might look to actually omit it

shipcat_definitions/src/structs/monitoring.rs
bh-anton-k: well it's the name for env key, and I think I've added a regex check for it to be only CAPS_SNAKE (not tested, yet)

shipcat_definitions/src/structs/monitoring.rs
bh-anton-k: there're problems with some use-cases when sentry is partially provisioned across some of environments defined. we might want to keep it enabled at service level and disable by overriding this property. deleting properties via `merge()` might be not supported, I have't yet grasped the macro magic

shipcat_filebacked/src/monitoring_source.rs
```diff
@@ -0,0 +1,390 @@
...
+///       params:
+///         duration: 60
+///         threshold: 0.5
+/// ```
+#[derive(Debug, Default, Clone, Deserialize)]
+#[cfg_attr(test, derive(PartialEq))]
+#[serde(deny_unknown_fields, rename_all = "camelCase")]
+pub struct MonitoringSource {
+    /// we might want to route all the alerts to some particular channel
+    pub slack: Option<SlackChannel>,
```
gleaver: Is this necessary? Feels a bit like a premature optimisation. We can already override per-alert, and this would be easy to add in later if it turns out to be a common use-case.

shipcat_filebacked/src/monitoring_source.rs
```diff
@@ -0,0 +1,390 @@
...
+        MonitoringSource {
+            slack: self.slack.merge(other.slack),
+            sentry: self.sentry.merge(other.sentry),
+            new_relic: merge_maps_deep(self.new_relic, other.new_relic, false),
+        }
+    }
+}
+
+impl Build<Monitoring, SlackChannel> for MonitoringSource {
+    fn build(self, squad_default_support: &SlackChannel) -> Result<Monitoring> {
```
gleaver: maybe just `default_channel` ?

shipcat_filebacked/src/monitoring_source.rs
```diff
@@ -0,0 +1,390 @@
...
+                name
+            ));
+        }
+
+        if self.template.is_none() {
+            bail!(format!("Template name for alert `{}` must be provided", name))
+        }
+
+        let template = self.template.unwrap();
+        let template_re = Regex::new(r"^[0-9a-zA-Z_\-]{1,50}$").unwrap();
```
gleaver: Error message says spaces are ok, but this regex doesn't

shipcat_definitions/src/manifest.rs
```diff
@@ -683,6 +683,34 @@ pub struct Manifest {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub rbac: Vec<Rbac>,
 
+    /// Monitoring section covering Sentry/NewRelic configurations
+    ///
+    /// ```yaml
+    /// monitoring:
```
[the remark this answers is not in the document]
bh-anton-k: this is post-merge format, not a fully templated one (it would be quite larger and include at least alert type and metric name)

shipcat_filebacked/src/monitoring_source.rs
bh-anton-k: for this to be easy to add I have to migrate the map under `/newRelic/alerts` - otherwise we won't have any injection points in the schema

shipcat_filebacked/src/monitoring_source.rs
bh-anton-k: error message for template below does not mention spaces, the one above does, but it has its own regex as well

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+    #[serde(default)]
+    pub alerts: BTreeMap<String, Option<NewRelicAlertSource>>,
+}
+
+impl Build<Option<NewRelic>, SlackChannel> for NewRelicSource {
+    fn build(self, default_channel: &SlackChannel) -> Result<Option<NewRelic>> {
+        let slack = self.slack.unwrap_or(default_channel.clone());
+        slack.verify()?;
+
+        if self.alerts.len() == 0 {
```
clux: .is_empty()

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+            return Ok(None);
+        }
+
+        let alerts_res: Result<BTreeMap<String, NewRelicAlert>> = self.alerts.into_iter()
+            .filter(|(_name, alert_source)| alert_source.is_some())
+            .map(|(name, alert_source)|
+                alert_source
```
clux: .filter_map(|(_, a)| a)

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+            )
+            .collect();
+
+        Ok(
+            Some(
+                NewRelic {
+                    alerts: alerts_res?,
+                }
+            )
+        )
```
clux: one line - unlist from rustfmt.toml

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+    pub template: Option<String>,
+    pub slack: Option<SlackChannel>,
+    #[serde(default)]
+    pub params: BTreeMap<String, String>,
+}
+
+impl Build<NewRelicAlert, (String, SlackChannel)> for NewRelicAlertSource {
+    fn build(self, name_slack: &(String, SlackChannel)) -> Result<NewRelicAlert> {
+        let (name, default_channel) = name_slack;
+        let slack = self.slack.unwrap_or(default_channel.clone());
```
clux: unwrap_or_else - run clippy

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+        let name_re = Regex::new(r"^[0-9a-zA-Z _\-]{1,50}$").unwrap();
+        if !name_re.is_match(name) {
+            bail!(format!(
+                "Please use alnum chars with dashes, underscores and spaces only for alert name, not `{}`",
+                name
+            ));
+        }
+
+        if self.template.is_none() {
+            bail!(format!("Template name for alert `{}` must be provided", name))
```
clux: redundant bail + format - more clippy

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+        }
+
+        let template = self.template.unwrap();
```
clux: if let Some(template) = self.template and bail in else

clux: or self.template.unwrap_or_else(|| bail!) or something like that

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+        let template_re = Regex::new(r"^[0-9a-zA-Z_\-]{1,50}$").unwrap();
+        if !template_re.is_match(&template) {
+            bail!(format!(
+                "Please use alnum chars with dashes, underscores only for template name for alert `{}`, not `{}`",
```
clux: shorten

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+                name, template
+            ));
+        }
+
+        Ok(
+            NewRelicAlert {
+                name: name.to_string(),
+                template: template,
+                slack: slack,
```
clux: shorthands

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+
+    #[derive(Serialize, Deserialize, Clone, Default)]
+    pub struct ManifestNarrowed {
+        #[serde(default, rename = "newRelic", skip_serializing_if = "Option::is_none")]
+        pub new_relic: Option<NewRelic>,
+    }
+
+    #[test]
+    fn merge_build_test() {
+        let test_data_table = [
```
clux: weird ad hoc vector of test data

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+            let manifest: ManifestOverridesNarrowed =
+                serde_yaml::from_str(manifest_yml)
+                    .expect(
+                        &format!("failed to parse main manifest for test `{}`", test)
+                    );
+            let prod: ManifestOverridesNarrowed =
+                serde_yaml::from_str(env_yml)
+                    .expect(
+                        &format!("failed to parse env manifest for test `{}`", test)
+                    );
```
clux: question mark?

shipcat_filebacked/src/new_relic_source.rs
```diff
@@ -0,0 +1,341 @@
...
+
+            let merge_with_env = manifest.merge(prod);
+
+            print!("merge result for test `{}`:\n{:#?}\n\n", test, merge_with_env);
```
clux: println

shipcat_filebacked/src/sentry_source.rs
```diff
@@ -0,0 +1,158 @@
...
+        if !dsn_env_name_re.is_match(&dsn_env_name) {
+            bail!(format!("Please use a valid env var name for DSN, not `{}`", dsn_env_name));
+        }
+
+        Ok(
+            Sentry {
+                slack: slack.clone(),
+                dsn_env_name,
+            }
+        )
```
clux: one line with shorthands

shipcat_filebacked/src/new_relic_source.rs
bh-anton-k: yes, this actually runs verify for both override and default cases, could be rephrased as 
```
        let slack = self.slack
            .map(SlackChannel::verify)
            .unwrap_or_else(Ok(default_channel.clone()))?;
```

shipcat_filebacked/src/new_relic_source.rs
bh-anton-k: ```self.template.ok_or(value_for_error)```

shipcat_filebacked/src/manifest.rs
```diff
@@ -154,7 +167,12 @@ impl Build<Manifest, (Config, Region)> for ManifestSource {
             kafka: kafka,
             sourceRanges: overrides.source_ranges.unwrap_or_default(),
             rbac: overrides.rbac.unwrap_or_default(),
-
+            newrelic: overrides.newrelic.build(&team_notifications)?,
+            sentry: if let Some(sentry_source) = overrides.sentry {
+                Some(sentry_source.build(&team_notifications)?)
+            } else {
+                None
+            },
```
clux: map

shipcat_filebacked/src/newrelic_source.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    #[test]
+    fn test_empty_all_the_things() -> Result<()> {
+        test_for(
+            r###"{}"###,
+            r###"{}"###,
+            "CDEFTEAM8",
+            r###"---
+{}"###,
+        )
```
clux: structify the inputs for test_for use String

clux: TestData { root: "ARSRAS".into(),
 override: "arstr".into()

shipcat_filebacked/src/newrelic_source.rs
```diff
@@ -0,0 +1,321 @@
...
+        env_yml: String,
+        default_channel: String,
+        crd_expected: String,
+    }
+
+    #[test]
+    fn test_empty_all_the_things() -> Result<()> {
+        test_for(TestSet {
+            manifest_yml: "{}".to_string(),
+            env_yml: "{}".to_string(),
```
clux: `.into()` looks cleaner in all of these but whatever

shipcat_filebacked/src/newrelic_source.rs
```diff
@@ -0,0 +1,321 @@
...
+      name: myApdex
+      template: SimpleApdex
+      slack: CDEFTEAM8
+      params:
+        priority: warning
+        threshold: "0.8""###.to_string(),
+        })
+    }
+
+    fn test_for(test_set: TestSet) -> Result<()> {
```
clux: test_merge is a more descriptive name here

shipcat_filebacked/src/sentry_source.rs
```diff
@@ -30,13 +36,17 @@ impl Build<Sentry, SlackChannel> for SentrySource {
             .map(|s| s.verify().map(|_| s))
             .unwrap_or_else(|| Ok(default_channel.clone()))?;
 
+        let notifications = self
+            .notifications
+            .unwrap_or(true);
```
clux: oh, that's how you are using it. that's also non-standard. default for missing bools is generally false.
`silent: false` default?

clux: then you could literally just do
```rust
#[serde(default)]
silent: bool
```

bh-anton-k: well, valid point, Guy and me were also a bit puzzled by defaulting None to true -- fixed

PR #310, opened 2019-10-11
shipcat restart

clux:
simple function to restart all pods safely.
requires deployment PATCH access.


shipcat_cli/src/main.rs
```diff
@@ -289,9 +289,18 @@ fn build_cli() -> App<'static, 'static> {
                     .help("Apply template even if no changes are detected"))
               .arg(Arg::with_name("service")
                 .required(true)
-                .help("Service to upgrad"))
+                .help("Service to apply"))
```
janslow: This has bugged me for so long 😄

clux: hahaha

PR #312, opened 2019-10-14
DEVP-1024 relying on yaml deletions instead of explicit enabled

bh-anton-k:


shipcat_filebacked/src/monitoring_source.rs
```diff
@@ -186,23 +193,23 @@ mod tests {
...
       slack: "#alert-level-channel-override"
-      enabled: true
       template: Default"###,
             //  REGIONAL OVERRIDES
             r###"---
-{}"###,
+monitoring:
+  newRelic:
+    all_params_are_default: {}
+    this_one_has_slack_defined_and_wont_propagate: {}"###,
```
clux: very liberal mixing of camelCase and snake_case here

shipcat_filebacked/src/monitoring_source.rs
```diff
@@ -32,17 +31,28 @@ pub struct MonitoringSource {
...
+        let sentry_merged_with_deletion =
+            if other.sentry.is_some() {
+                if self.sentry.is_some() {
+                    Some(self.sentry.unwrap().merge(other.sentry.unwrap()))
+                } else {
+                    self.sentry.merge(other.sentry)
+                }
+            } else {
+                None
+            };
```
clux: this is pretty much caused by sentry being nested under monitoringsource, right?
is it problematic having it straight on the Manifest?

shipcat_filebacked/src/monitoring_source.rs
```diff
@@ -32,17 +31,28 @@ pub struct MonitoringSource {
...
+
         MonitoringSource {
             slack: self.slack.merge(other.slack),
-            sentry: self.sentry.merge(other.sentry),
-            new_relic: merge_maps_deep(self.new_relic, other.new_relic, false),
+            sentry: sentry_merged_with_deletion,
+            new_relic: merge_maps_deep(self.new_relic, other.new_relic, true),
```
clux: same question

shipcat_filebacked/src/monitoring_source.rs
gleaver: I think these are arbitrary string names - they could be `"this one has slack defined": {}`

shipcat_filebacked/src/monitoring_source.rs
bh-anton-k: If we move this directly to manifest level then I won't be able to weave this custom merge behaviour in.
...which means that the last test case (disabling sentry with a missing yaml subtree for one region while having it on by default) will break. 

We need this custom merge because default one for `Option<T>` does not work towards this usecase: missing subtrees, explicit `null` or `~` value in yaml is going to deserialise to `None` in the overrides `None`, which has no effect over defined main value after `merge()`.

So to move it and/or to eradicate this custom merge impl we should contemplate giving up on hole-punching behaviour for sentry (which I think is totally valid as well).

shipcat_filebacked/src/monitoring_source.rs
bh-anton-k: Moving won't help it, I think. Here intended behaviour is indeed a bit more complex than usual one-level map merge (we want to override properties and merge nested `params` fields so that we might change thresholds or levels while keeping other template parameter values as they were defined in the original yaml). This [testcase](https://github.com/Babylonpartners/shipcat/blob/a340108716ae194dd93df356b82e38e77b8f358b/shipcat_filebacked/src/monitoring_source.rs#L263-L309) covers an example.

As sentry section is rather small, yaml-wise, the amount of copypaste is tolerable if we give up on hole-punching behaviour (the code above). For template and params (this code) we might provoke quite heftier amounts of copy-paste so I think it's better to retain this implementation.

shipcat_filebacked/src/monitoring_source.rs
bh-anton-k: now fixed in main PR and merged into this one as well

shipcat_filebacked/src/monitoring_source.rs
bh-anton-k: https://github.com/Babylonpartners/shipcat/pull/312/commits/f7058c8800eb04df73449a80df258acda6e55e12  resolves this

shipcat_filebacked/src/monitoring_source.rs
bh-anton-k: https://github.com/Babylonpartners/shipcat/pull/312/commits/ccacebef8ee3f534e88a3eb48e06e1af8c91b5a6 resolves this

PR #314, opened 2019-10-16
Diph 138 eventstream shipcat values

bclewis1:
This exposes a new struct `eventStreams` in the values shipcat produces.

If I add the following to my `manifest.yml`:
```
eventStreams:
  events.signup.registered:
    bufferWorker:
      workerType: default
      env:
        DATABASE_CONN_STRING: "{{ parent_env.POSTGRES_HOST }}"
```

I get the following in my `shipcat values`
```
...
eventStreams:
  events.signup.registered:
    buffer_worker:
      worker_type: default
      parent_env:
        plain:
          (the core-auth-2 environment)
      env:
        plain:
          DATABASE_CONN_STRING: "{{ parent_env.POSTGRES_HOST }}"
...
```

The idea would then be to chart out worker pods for each of the eventStreams registered in the `eventStreams` map.


fnobilia:
@bclewis1 the more I think about this approach and the more I am not convinced. In the medium term, I would expect a structure like that for `event-streams`
```
event_streams:
  events.appointments_booked:
    event_definitions:
      - key: 'ProtoKeyX'
        value: 'ProtoValueX'
    allowed_consumers: ["user-service", "bi"]
    allowed_producers: ["user-service"]
```
The `event-flusher` should be outside this and list together with other workers to have a more clean separation of concern.

bclewis1:
okay but then we are right back to Guy's objection that we can't have the version of a worker float separately - it needs to be maintained globally. If you aren't convinced, whats your alternative? I'm happy to take whatever approach.

clux:
Succeeded by #381

PR #318, opened 2019-10-16
Support multiple Kong APIs per service

janslow:
Currently, service which need to vary Kong configuration for different paths must use the per region `extra_apis` in `shipcat.conf` to declare them.

This requires them to manually maintain the configuration for each region and manually declare values which are usually defaulted (such as `upstream_url`).

This PR allows a `kongApis` property to be used instead of `kong` (mutually exclusive).
* `kongApis` is a map of API name to API configuration.
* API names must either match the service or use that as a prefix, to make ownership of APIs obvious.
* A special `defaults` key is used to allow configuration to be shared across multiple APIs, if desired.

See https://github.com/Babylonpartners/manifests/pull/15512/files for an example PR.


shipcat_filebacked/src/kong.rs
```diff
@@ -2,10 +2,68 @@ use merge::Merge;
...
+    pub kong: KongConfig,
+    // TODO: Remove Manifest.kong
+    pub single_api: Enabled<KongSource>,
+}
+
+impl Build<Vec<Kong>, KongApisBuildParams> for KongApisSource {
+    fn build(self, params: &KongApisBuildParams) -> Result<Vec<Kong>> {
+        let defaults = Enabled { enabled: None, item: self.defaults };
+
+        if let Some(Some(k)) = defaults.clone().merge(params.single_api.clone()).build(&KongBuildParams {
```
clux: what's with the awkward double Option here?

shipcat_filebacked/src/kong.rs
```diff
@@ -2,10 +2,68 @@ use merge::Merge;
...
+                name,
+                service: params.service.clone(),
+                region: params.region.clone(),
+                kong: params.kong.clone(),
+            })?;
+            if let Some(Some(api)) = maybe {
+                built.push(api);
+            }
+        }
+        Ok(built)
```
clux: this last part looks like a `filter_map` on `self.apis` - but suppose you're doing some error handling in there

shipcat_filebacked/src/kong.rs
```diff
@@ -31,55 +89,55 @@ pub struct KongSource {
...
+    pub kong: KongConfig,
 }
 
 impl Build<Option<Kong>, KongBuildParams> for KongSource {
     /// Build a Kong from a KongSource, validating and mutating properties.
     fn build(self, params: &KongBuildParams) -> Result<Option<Kong>> {
-        let KongBuildParams { region, service } = params;
-        if let Some(k) = &region.kong {
-            let hosts = self.build_hosts(&k.base_url)?;
+        let KongBuildParams { region, service, name, kong } = params;
```
clux: wait you got rid of the check? does this work in ops cluster?

shipcat_filebacked/Cargo.toml
```diff
@@ -15,3 +15,4 @@ serde_yaml = "0.8.9"
 log = "0.4.5"
 error-chain = "0.12.0"
 walkdir = { version = "2.2.5"}
+maplit = "1.0.2"
```
clux: dev-dependencies?

shipcat_filebacked/src/kong.rs
```diff
@@ -31,55 +89,55 @@ pub struct KongSource {
...
+        debug!("Building Kong API {} for {}", &name, &service);
 
-            if hosts.is_empty() && self.uris.is_none() {
-                return Ok(None);
-            }
+        let hosts = self.build_hosts(&kong.base_url)?;
 
-            let upstream_url = self.build_upstream_url(&service, &region.namespace);
-            let (auth, authorization) = KongSource::build_auth(self.auth, self.authorization)?;
+        if hosts.is_empty() && self.uris.is_none() {
```
clux: does this work now? kong.hosts might not be set anymore after my last merge

shipcat_filebacked/src/manifest.rs
```diff
@@ -171,11 +173,13 @@ impl ManifestSource {
 
         let overrides = self.overrides.clone();
         let defaults = overrides.defaults;
-        let kong_apis = if region.kong.is_some() {
-            defaults.kong.build(&KongBuildParams {
+        let kong_apis = if let Some(k) = &region.kong {
```
clux: ah you're doing the check here :+1:

shipcat_filebacked/src/manifest.rs
```diff
@@ -171,11 +173,13 @@ impl ManifestSource {
...
+            defaults.kong_apis.build(&KongApisBuildParams {
                 service: base.name.to_string(),
                 region: region.clone(),
-            })?.unwrap_or_default().map(|k| vec![k]).unwrap_or_default()
+                kong: k.clone(),
+                single_api: defaults.kong,
+            })?
         } else {
             // NB: this drops kong entries on the floor if region.kong is None
```
clux: good.

shipcat_filebacked/src/util/enabled.rs
```diff
@@ -24,14 +43,72 @@ impl<S: Build<B, P> + Merge, B, P> Build<Option<B>, P> for Enabled<S> {
     }
 }
 
+/// EnabledMap is a map where each value is wrapped in an Enabled.
+///
+/// It can be built into a map which flattens the Enabled wrappers, so disabled values are excluded.
+#[derive(Deserialize, Default, Clone, PartialEq)]
+#[cfg_attr(test, derive(Debug))]
+pub struct EnabledMap<K: Clone + std::hash::Hash + Ord, V: Clone + Default + Merge>(BTreeMap<K, Enabled<V>>);
```
clux: ah this must be where they are hiding the good code

shipcat_filebacked/src/util/enabled.rs
```diff
@@ -24,14 +43,72 @@ impl<S: Build<B, P> + Merge, B, P> Build<Option<B>, P> for Enabled<S> {
...
+    type IntoIter = std::collections::btree_map::IntoIter<K, Enabled<V>>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.0.into_iter()
+    }
+}
+
+impl<K, VS, VB, P> Build<BTreeMap<K, VB>, P> for EnabledMap<K, VS> where
+    K: Clone + std::hash::Hash + Ord,
+    VS: Build<VB, P> + Clone + Default + Merge {
```
clux: hahahah. omg.

shipcat_filebacked/src/kong.rs
janslow: To maintain the legacy behaviour, you can also disable Kong by setting `uri` and `hosts` to empty, in addition to the new `enabled: false`.

shipcat_filebacked/src/kong.rs
janslow: Yep, I'm not really aware of a nice way to use `Result` in an iterator.

shipcat_filebacked/src/kong.rs
janslow: I realised you'd already done a check that `region.kong` was `Some`, so I just pass the real kong down instead.

shipcat_filebacked/src/kong.rs
clux: collect() returns the first one!

shipcat_filebacked/src/kong.rs
janslow: I've pushed this down a bit, so it only affects `kong` for legacy behaviour.

After this PR is live, the APIs relying on this legacy behaviour can be migrated to `enabled: false` (then the legacy behaviour can be dropped).

shipcat_filebacked/src/util/enabled.rs
janslow: Yeah, this makes me sad.

I really wish Rust didn't force you to redeclare the bounds defined on the type declaration at every impl.

Maybe https://github.com/nox/rust-rfcs/blob/master/text/0213-defaulted-type-params.md will help

shipcat_filebacked/src/kong.rs
janslow: Oh, yeah. I'd forgotten about that.

I still couldn't get a nice solution, because I'm trying to convert `Iterator<Result<Option<Kong>>>` to `Iterator<Option<Result<Kong>>>` to `Iterator<Result<Kong>>` to `Result<Vec<Kong>>`.

The best I've got is as follows, but it has a weird `filter_map` (to convert `Result<Option<T>>` to `Option<Result<T>>`)

```rs
        self.apis.into_iter().map(|(name, k)| {
            debug!("Building Kong API {}", &name);
            if name != params.service && !name.starts_with(&format!("{}-", &params.service)) {
                // TODO: Duplicate Kong's name validation
                bail!("Kong API name must be '${SERVICE_NAME}' or '${SERVICE_NAME}-*'")
            }
            let merged = defaults.clone().merge(k);
            merged.build(&KongBuildParams {
                name,
                service: params.service.clone(),
                region: params.region.clone(),
                kong: params.kong.clone(),
            })
        })
        .filter_map(|r| match r {
            Result::Ok(None) => None,
            Result::Ok(Some(x)) => Some(Ok(x)),
            Result::Err(e) => Some(Err(e)),
        }).collect()
```

clux: sounds like you might be able to use `transpose` here, but ultimately it'll probably just make it hard to follow

janslow: Agreed, I'll leave it as a loop

PR #320, opened 2019-10-17
Expose all ports

jackatbancast:
This PR adds the ability to port-forward to every port listed in the manifest.

There is an edge case that is not currently handled, where if a port below 1024 is used, and a port of `7778` is used, then both will be allocated the same port on the local machine.
The underlying `kubectl` call should catch this and error. As such no effort has been made to fix this case yet.


shipcat_cli/src/kubectl.rs
```diff
@@ -457,26 +457,36 @@ pub fn shell(mf: &Manifest, desiredpod: Option<usize>, cmd: Option<Vec<&str>>) -
...
     // TODO: kubectl auth can-i create pods/portforward first
-    let port = if let Some(p) = mf.httpPort {
-        p
-    } else if let Some(p) = mf.ports.iter().find(|p| p.name == "http") {
-        p.port
-    } else if let Some(p) = mf.ports.iter().find(|p| p.name == "grpc") {
-        p.port
-    } else {
-        bail!("{} does not expose an http/grpc port to port-forward to", mf.name)
+    let mut port_count = 7777;
```
clux: port_range_lower, unmut

shipcat_cli/src/kubectl.rs
```diff
@@ -457,26 +457,36 @@ pub fn shell(mf: &Manifest, desiredpod: Option<usize>, cmd: Option<Vec<&str>>) -
...
+    let mut ps: Vec<_> = mf.ports.iter().map(|mp| mp.port).collect();
+
+    if let Some(p) = mf.httpPort {
+        if ps.iter().find(|&&mp| mp == p).is_none() {
+            ps.push(p);
+        }
+    };
```
clux: so ports is the union of mf.ports and mf.httpPort.

shipcat_cli/src/kubectl.rs
```diff
@@ -457,26 +457,36 @@ pub fn shell(mf: &Manifest, desiredpod: Option<usize>, cmd: Option<Vec<&str>>) -
...
+
+    if ps.len() == 0 {
+        bail!("{} does not expose any port to port-forward to", mf.name)
```
clux: i guess that's an error, even though you've port-forwarded all ports that exist :thinking:

shipcat_cli/src/kubectl.rs
```diff
@@ -457,26 +457,36 @@ pub fn shell(mf: &Manifest, desiredpod: Option<usize>, cmd: Option<Vec<&str>>) -
...
     };
-    // first 1024 ports need sudo so avoid that
-    let localport = if port <= 1024 { 7777 } else { port };
 
-    debug!("Port forwarding kube deployment {} to localhost:{}", mf.name, localport);
+    let ports = ps.iter().map(|&port| {
+        let localport = if port <= 1024 { port_count += 1; port_count } else { port };
```
clux: rather than mutating and index, use `Iterator::enumerate` to get it and add it to the lower bound

shipcat_cli/src/kubectl.rs
```diff
@@ -457,26 +457,36 @@ pub fn shell(mf: &Manifest, desiredpod: Option<usize>, cmd: Option<Vec<&str>>) -
 /// Useful because we have autocomplete on manifest names in shipcat
 pub fn port_forward(mf: &Manifest) -> Result<()> {
     // TODO: kubectl auth can-i create pods/portforward first
```
clux: wanna do this shell out as well?

clux:
also remove the version bump, other version en route

shipcat_cli/src/kubectl.rs
jackatbancast: I don't really want to add it in this PR. It doesn't really fit in the scope of the changes.

shipcat_cli/src/kubectl.rs
jackatbancast: Whilst technically it is port-forwarding all that exists, I thought it would be more helpful to have an error indicating that nothing has actually happened.

shipcat_cli/src/kubectl.rs
janslow: Yeah, there's no real reason why `httpPort` can't be merged into the `ports` array earlier on, after we've removed the special logic in manifests (we need to introduce the concept of the `servicePort` field to `Port`, i.e., the port it's exposed as on the service).

shipcat_cli/src/kubectl.rs
clux: fair

shipcat_cli/src/kubectl.rs
jackatbancast: @janslow in that case should we just remove `httpPort` and list the port in `ports` as a `http` port?

janslow: As discussed, we can't do this yet.

PR #325, opened 2019-10-18
Use Kong API name not service name in `get apistatus`

### LINKED ISSUE #127: Support multiple Kong APIs per service

janslow:
#318 doesn't currently work for external APIs.

Also, fixes #127


clux:
doh.

PR #326, opened 2019-10-21
[OB-61] kubernetes 1.15.5 client

clux:
needed for rollout restart


Dockerfile.circleci
```diff
@@ -1,6 +1,6 @@
 FROM circleci/python:3.7.4-node
 
-ENV KUBEVER=1.14.6 \
+ENV KUBEVER=1.15.3 \
```
janslow: Are these intentionally different?

clux: nope, fixed.

PR #327, opened 2019-10-21
Add can-i helper and wrap port_forward

jackatbancast:
This PR introduces a function that allows us to check if we have authorization to perform an action on a resource.
The first use-case of this is reporting the inability to port-forward based on priviledges of your token.


shipcat_cli/src/kubectl.rs
```diff
@@ -456,6 +471,10 @@ pub fn shell(mf: &Manifest, desiredpod: Option<usize>, cmd: Option<Vec<&str>>) -
 ///
 /// Useful because we have autocomplete on manifest names in shipcat
 pub fn port_forward(mf: &Manifest) -> Result<()> {
+    if kani(mf, "create".into(), "pods/portforward".into())? != true {
+        bail!("Current token does not have authorization to port-forward")
+    };
```
clux: i don't think you should be doing the call here, as noted in the issue - this style of pattern will just pointlessly double our requests if taken to the extreme.

shipcat_cli/src/kubectl.rs
```diff
@@ -31,6 +31,21 @@ fn kout(args: Vec<String>) -> Result<(String, bool)> {
...
+        "can-i".into(),
+        verb,
+        resource,
+    ];
+    use std::process::Command;
+    debug!("kubectl {}", args.join(" "));
+    let s = Command::new("kubectl").args(&args).output()?;
+    Ok(s.status.success())
+
+}
```
clux: reasonable :+1:

shipcat_cli/src/kubectl.rs
clux: although.. maybe we don't need to take it to an extreme. we could just be doing this style of thing as sanity for the non-ci operations :thinking:

jackatbancast: In this case I think it'd probably be fine as we shouldn't be port forwarding in CI.

But as a general case/pattern it may be useful to hide behind a flag?
That way we could disable it in CI, and give better errors to end users.

clux: yeah, that could potentially be useful. maybe a global strict flag like how the version check is done perhaps?

Issue #262, opened 2019-08-15
better upgrade solution


clux:
Now that shipcat is kind of essential within babylon, it'd be good to have an easier way for developers to upgrade. Something like what [lal upgrade](https://github.com/cisco/lal-build-manager/blob/master/src/upgrade.rs) had would be nice. Autoupgrading is not necessarily nice (bypassing the package manager), but we can shoe-horn it into a feature that's used within babylon.

Alternatively, we could come up with a way in `shipcat.conf` to not force users to upgrade between non-breaking changes.

clux:
https://github.com/jaemk/self_update is a thing. Maybe we can do a compile time feature for it.

clux:
goofing around a bit with this on the on the `self-upgrade` branch

PR #328, opened 2019-10-22
auto-upgrade

clux:
For #262 

Basically exposes a `shipcat self-upgrade` command. This will fetch a specified/latest tag via github releases, unpack the tarball in a temp directory, and swap over the binary in a 3-way swap (potentially screwing up a `brew` installation). It will even swap out the binary at the symlink position if it's linked to somewhere (so people building locally might find themselves with a downloaded release at  `./target/debug/shipcat` - which is probably fine (confused usage)).

When `SHIPCAT_AUTOUPGRADE` evar is set, `shipcat` will invoke `self-upgrade` internally when the version found in shipcat.conf is newer than what we have.

Not tested for Darwin. Need eyes on this please.


clux:
Because github api rate limits from a single ip, i've added the option to read a github api token optionally under `SHIPCAT_AUTOUPGRADE_TOKEN`. It can have the same permissions as a brew token (repo read).

shipcat_cli/src/main.rs
```diff
@@ -465,8 +472,17 @@ fn resolve_config(args: &ArgMatches, ct: ConfigType) -> Result<(Config, Region)>
         if args.is_present("strict-version-check") {
             return Err(e.into())
         } else {
-            warn!("shipcat version less than pinned minimum - results may vary");
-            warn!("{}", e);
+            if let Ok(_) = std::env::var("SHIPCAT_AUTOUPGRADE") {
```
janslow: This can be flattened into an `else if let Ok(_) = std::env::var("SHIPCAT_AUTOUPGRADE")`

clux: Ah, thanks. Had forgotten.

clux:
Have tested this with brew. We need to tell people to uninstall it from brew for this to work. Any `brew upgrade shipcat` calls will fail because we overwrite the symlink `brew` manages.

clux:
Merging it for now. It's a feature flagged thing anyway.

PR #334, opened 2019-10-23
[IDP-20] Remove `anonymous_consumers` from region configuration

janslow:
It's not used internally any more. There's a hardcoded `anonymous` consumer instead.


janslow:
* [ ] Blocked on https://github.com/babylonhealth/manifests/pull/15886

PR #335, opened 2019-10-23
[IDP-20] Remove support for Kong OAuth plugin

janslow:


janslow:
* [ ] Blocked on https://github.com/babylonhealth/manifests/pull/15889
* [ ] Blocked on https://github.com/babylonhealth/manifests/pull/15888

shipcat_filebacked/src/kong.rs
```diff
@@ -165,20 +165,16 @@ impl KongSource {
...
-            }
-            (Some(_), Some(_)) => bail!("auth must be unset or JWT if authorization is enabled"),
-            // otherwise
-            (Some(x), _) => Ok((x.clone(), None)),
-            (None, _) => Ok((Authentication::default(), None)),
-        }
+            (None, Some(a)) => (None, Some(a)),
+            (Some(_), Some(_)) => bail!("auth and authorization.enabled are mutually exclusive"),
+            (Some(Authentication::None), None) | (None, None) => (None, None),
+            (Some(Authentication::Jwt), None) => (Some(Authentication::Jwt), None),
```
clux: i fell like this is one error case in the (Some, Some) case and the three other cases can be a single case without excessive pattern matching

janslow: I've simplified it further.

However, there are still three cases, because `Some(Authentication::None)` is now equivalent to `None`.

i.e., instead of 

```yaml
kong:
  authorization:
    enabled: false
  auth: none
```

you can now do:

```yaml
kong:
  authorization:
    enabled: false
```

shipcat_definitions/src/region.rs
```diff
@@ -238,16 +239,6 @@ pub struct KongJwtConsumer {
...
 
-impl KongConfig {
-    fn secrets(&mut self, vault: &Vault, region: &str) -> Result<()> {
-        if self.oauth_provision_key == "IN_VAULT" {
-            let vkey = format!("{}/kong/oauth_provision_key", region);
-            self.oauth_provision_key = vault.read(&vkey)?;
-        }
-        Ok(())
-    }
-}
```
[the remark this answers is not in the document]
janslow: Yep, I'm cleaning up `shipcat kong` here https://github.com/babylonhealth/shipcat/pull/338

PR #337, opened 2019-10-23
[IDP-20] Don't use secrets in `shipcat kong`

janslow:
After https://github.com/babylonhealth/shipcat/pull/335 and https://github.com/babylonhealth/shipcat/pull/333, the Kongfig output no longer contains any secrets.

Therefore don't require `shipcat kong` to use the manifests with secrets. This will allow us to diff Kong configuration in Danger.js


janslow:
I didn't intend to create a non-draft PR yet. Replaced by https://github.com/babylonhealth/shipcat/pull/338

PR #347, opened 2019-11-06
[OB-114] shipcat top

clux:


clux:
Have documented the analytical vs dynamic nature of this implementation of `top` a bit better within. Doing this dynamically is not really within the purview of shipcat anyway so keeping this as is.

PR #348, opened 2019-11-07
[OB-114] shipcat top --world

clux:


shipcat_cli/src/top.rs
```diff
@@ -52,7 +94,15 @@ pub fn resources(order: ResourceOrder, upper_bounds: bool, conf: &Config, reg: &
...
-    let mut mfs = mfs_res?;
+    let mfs = sort_and_print_resources(mfs_res?, order, upper_bounds)?;
+    Ok(mfs)
+}
+
+
+fn sort_and_print_resources(mfs_: Vec<(Manifest, ResourceTotals)>, order: ResourceOrder, upper_bounds: bool)
+    -> Result<Vec<(Manifest, ResourceTotals)>>
+{
+    let mut mfs = mfs_;
```
skrytt: what's the intent here?

clux: it's a move into a mutable var so `sort_by` can change order. tried to do it in the signature, but the compiler didn't like it :man_shrugging:

PR #349, opened 2019-11-07
HANDS-35

chris-denman-babylon:
Hub & Spoke service routing functionality.


shipcat_definitions/src/config.rs
```diff
@@ -26,16 +25,19 @@ pub struct ManifestDefaults {
     /// Chart to defer to
     pub chart: String,
     /// Default replication counts
-    pub replicaCount: u32
-
+    pub replicaCount: u32,
+    /// Default region_id_pattern
+    pub region_id_pattern: String,
```
clux: camelCase please

shipcat_definitions/src/manifest.rs
```diff
@@ -359,6 +360,10 @@ pub struct Manifest {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub dependencies: Vec<Dependency>,
 
+    /// Service Routing Information
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub serviceRouting: Option<Vec<ServiceRouting>>,
```
clux: can we perhaps have a different name? `service` is implied already due to it being a shipcat service, and `Routing` is super general.

shipcat_definitions/src/structs/servicerouting.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn is_ok(region_id: &str, destination_host: &str, pattern: &str) {
+        assert!(
+            build_serviceRouting(region_id, destination_host)
+                .verify(pattern).is_ok());
+    }
+
+    fn is_err(region_id: &str, destination_host: &str, pattern: &str) {
+        assert!(
+            build_serviceRouting(region_id, destination_host)
+                .verify(pattern).is_err());
```
clux: unwrap_err for these things is better than assert(is_err()) because it will dump a helpful backtrace of where it went frong

shipcat_definitions/src/structs/servicerouting.rs
```diff
@@ -0,0 +1,120 @@
...
+        return ServiceRouting {
+            region_id: String::from(region_id),
+            destination_host: String::from(destination_host)
+        }
+    }
+
+    fn is_ok(region_id: &str, destination_host: &str, pattern: &str) {
+        assert!(
+            build_serviceRouting(region_id, destination_host)
+                .verify(pattern).is_ok());
```
clux: unwrap similarly is better than `assert(is_ok())`

shipcat_definitions/src/structs/servicerouting.rs
```diff
@@ -0,0 +1,120 @@
...
+        if !ServiceRouting::matches(&self.region_id,
+                                    &defaultRegionIdPattern) {
+            erroneous = true;
+            error!("Service Routing: Region ID \"{}\" is invalid.",
+                   &self.region_id);
+        }
+
+        if !ServiceRouting::is_dns_label(&self,&self.destination_host) {
+            erroneous = true;
+            error!("Service Routing: Destination host \"{}\" is not a valid DNS label.",
```
clux: no need to prefix error strings, that's already done by the logger. keep them short.

shipcat_definitions/src/structs/servicerouting.rs
```diff
@@ -0,0 +1,120 @@
...
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct ServiceRouting {
+    pub region_id: String,
+    pub destination_host: String
+}
+
+impl ServiceRouting {
+
+    pub fn matches(candidate: &str, pattern: &str) -> bool {
```
clux: does not need to expose this. in fact, having this method here just looses type safety by having two str arguments.

shipcat_definitions/src/structs/servicerouting.rs
```diff
@@ -0,0 +1,120 @@
...
+const DNS_RFC1123_LABEL_MAX_LENGTH: u8 = 63;
+
+// https://github.com/istio/istio/blob/6c96552ed3954a89989b12f35c2bb91bb5e36b6f/pkg/config/labels/instance.go#L95.
+const DNS_RFC1123_LABEL_PATTERN: &'static str = "^(?:[A-Za-z0-9][-A-Za-z0-9_\\.]*)?[A-Za-z0-9]$";
+
+#[derive(Serialize, Deserialize, Clone)]
+pub struct ServiceRouting {
+    pub region_id: String,
+    pub destination_host: String
+}
```
clux: please heavily docstring the intent behind these fields.

shipcat_definitions/src/manifest.rs
```diff
@@ -359,6 +360,10 @@ pub struct Manifest {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub dependencies: Vec<Dependency>,
 
+    /// Service Routing Information
```
clux: Needs a lot of documentation here.

shipcat_definitions/src/config.rs
chris-denman-babylon: Done

shipcat_definitions/src/manifest.rs
chris-denman-babylon: OK, i'll take it to the squad.

shipcat_definitions/src/structs/servicerouting.rs
chris-denman-babylon: Thanks! I'll read up and apply it when I understand...

chris-denman-babylon: Lovely stuff, thanks for the suggestion.

shipcat_definitions/src/structs/servicerouting.rs
chris-denman-babylon: Thx

shipcat_definitions/src/manifest.rs
alexfanshawebabylon: how about `Redirection`?

chris-denman-babylon: I think "Routing" is better than "Redirection", how about something like Region(al)Routing, SpokeRouting, Region(Request)Hosts, RegionHostMappings... shall think more. It's a quick fix regardless.

chris-denman-babylon: RegionalServices

alexfanshawebabylon: RegionalRouting?

alexfanshawebabylon: HeaderRedirect

alexfanshawebabylon: HeaderRouting

chris-denman-babylon: I think this:

```yaml
routes:
  -  region: "GB-UK"
      name:	‘profiles.dev-uk.local’
```

, is the winner. Comments welcomed.

janslow: Can we have some proper examples of how it's going to look. E.g., a draft PR in `manifests` with full examples

shipcat_definitions/src/config.rs
```diff
@@ -228,6 +230,10 @@ impl Config {
             bail!("image prefix must not end with a slash");
         }
 
+        if defs.regionPattern.trim().is_empty() {
```
alexfanshawebabylon: Can we keep this optional, only validate when we have a destination section

shipcat_definitions/src/manifest.rs
```diff
@@ -359,7 +360,22 @@ pub struct Manifest {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub dependencies: Vec<Dependency>,
 
-    /// Worker `Deployment` objects to additinally include
+    /// Destination Rules
+    ///
+    /// A sequence of DestinationRule structures.
+    ///
+    /// Used to forward requests to alternative (regional) implementations if they posses regional
+    /// identifiers that match the values given here.
```
clux: This is super vague. How will the forwarding occur?

shipcat_definitions/src/structs/destinationrule.rs
```diff
@@ -0,0 +1,103 @@
+use regex::Regex;
+use super::Result;
+
+/// DestinationRule
+///
+/// An abstraction for forwarding service-requests to alternative implementations.
+///
+/// Incoming service requests will be examined (in a implementation defined way) to see if they
+/// possess a region  equal to that given by this structure.
```
clux: Implementation defined way, but given the code, it's heavily implied to be istio. Can we have an example PR in manifests to see how this would be consumed before accepting this? There might be better ways than cramming experimental syntax in as a "standard" to be use for half a thousand microservices.

shipcat_definitions/src/config.rs
chris-denman-babylon: Done

shipcat_definitions/src/manifest.rs
alexfanshawebabylon: Is it for shipcat to know that?
I rather thought this is a combiner of config not implementation?

shipcat_definitions/src/structs/destinationrule.rs
alexfanshawebabylon: manifest PR: https://github.com/babylonhealth/manifests/pull/16685

shipcat_definitions/src/manifest.rs
clux: It doesn't necessarily need to, I just see other references to istio, and am wondering how disconnected it really is.

shipcat_definitions/src/structs/destinationrule.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    pub fn verify(&self, regionPattern: &str) -> Result<bool> {
+        let mut erroneous = false;
+
+        if regionPattern.trim().is_empty() {
+            bail!("region pattern not specified");
+        }
+
+        if !Regex::new(&regionPattern).unwrap().is_match(&self.region) {
```
clux: this will just bomb out here when the regex is invalid

shipcat_definitions/src/structs/destinationrule.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn is_dns_label(value: &str) -> bool {
+        value.len() <= (DestinationRule::DNS_LABEL_MAX_LENGTH as usize) &&
+            Regex::new(DestinationRule::DNS_LABEL_PATTERN).unwrap().is_match(value)
+    }
+
+    pub fn verify(&self, regionPattern: &str) -> Result<bool> {
+        let mut erroneous = false;
+
+        if regionPattern.trim().is_empty() {
+            bail!("region pattern not specified");
```
clux: will this not force us to put a region pattern everywhere up front? i'd rather have it opt in when possible.

shipcat_definitions/src/structs/destinationrule.rs
chris-denman-babylon: Fixed, we're serde-ing to Regex now.

shipcat_definitions/src/structs/destinationrule.rs
chris-denman-babylon: After talking to Eirik, we've put it in region config now.

shipcat_definitions/src/structs/servicerouting.rs
chris-denman-babylon: Done

shipcat_definitions/src/manifest.rs
```diff
@@ -778,6 +793,25 @@ impl Manifest {
...
+    pub fn verify_some_destination_rules(&self, region: &Region) -> Result<()> {
+        if let Some(ref _destinationRules) = &self.destinationRules {
+            match &region.destinationRuleRegionRegex {
+                Some(destinationRuleRegionRegex) => {
+                    for dr in _destinationRules {
+                        dr.verify(destinationRuleRegionRegex)?;
+                    }
+                },
+                None => bail!(
+                "Cannot verify destinationRules without a destinationRuleRegionRegex definition."),
```
clux: This message might not going to be helpful to users. How about:

```
Cannot use `destinationRules` in a region without a `destinationRuleRegex`
```

shipcat_definitions/src/manifest.rs
```diff
@@ -778,6 +793,25 @@ impl Manifest {
         Ok(self)
     }
 
+    /// Verifies the "destinationRules" manifest entries if they are configured
+    ///
+    /// It is erroneous to define destination rules without configuring the corresponding region's
+    /// destination rules region regular expression
+    pub fn verify_some_destination_rules(&self, region: &Region) -> Result<()> {
+        if let Some(ref _destinationRules) = &self.destinationRules {
+            match &region.destinationRuleRegionRegex {
```
clux: this match is an `if let Some(dr) = region.destinationRuleRegex` and a bail in the else clause

shipcat_definitions/src/manifest.rs
```diff
@@ -778,6 +793,25 @@ impl Manifest {
         Ok(self)
     }
 
+    /// Verifies the "destinationRules" manifest entries if they are configured
+    ///
+    /// It is erroneous to define destination rules without configuring the corresponding region's
+    /// destination rules region regular expression
+    pub fn verify_some_destination_rules(&self, region: &Region) -> Result<()> {
```
clux: `verify_regional_destination_rule` it's not just checking if it's a `Some`

shipcat_definitions/src/region.rs
```diff
@@ -521,6 +522,9 @@ pub struct Region {
     /// Default values for services
     #[serde(skip_serializing, default)]
     pub defaults: DefaultConfig,
+    /// The regular expression used to verify destination rules' regions
+    #[serde(default, skip_serializing_if = "Option::is_none", with = "serde_regex")]
+    pub destinationRuleRegionRegex: Option<Regex>,
```
clux: this does introduce a new dependency in https://github.com/tailhook/serde-regex/blob/master/src/lib.rs - which is a bit heavier than just calling .expect() on the string to convert it to a regex, but it's a pretty elegant solution.

shipcat_definitions/src/structs/destinationrule.rs
```diff
@@ -0,0 +1,95 @@
...
+use super::Result;
+
+/// DestinationRule
+///
+/// An abstraction that captures the information needed to make regional routing decisions.
+#[derive(Serialize, Deserialize, Clone)]
+pub struct DestinationRule {
+
+    /// The region the incoming request must possess to be considered for forwarding
+    pub region: String,
```
clux: this is overloading the term `region` within shipcat. it does not appear like it needs to match a region.name

shipcat_definitions/src/structs/destinationrule.rs
```diff
@@ -0,0 +1,95 @@
...
+    const MATCH_NONE_PATTERN: &'static str = ".^";
+    const DESTINATION_HOST_VALID: &'static str = "hostname.local";
+    const DESTINATION_HOST_INVALID: &'static str = ">:(";
+    const REGION: &'static str = "REGION";
+
+    fn withRegionAndDestinationHost(region: &str, destination_host: &str) -> DestinationRule {
+        return DestinationRule {
+            region: String::from(region),
+            destinationHost: String::from(destination_host)
+        }
```
clux: don't need `return` here, and you can just do `region.into()` rather than the big constructor call

shipcat_definitions/src/structs/destinationrule.rs
```diff
@@ -0,0 +1,95 @@
...
+///
+/// An abstraction that captures the information needed to make regional routing decisions.
+#[derive(Serialize, Deserialize, Clone)]
+pub struct DestinationRule {
+
+    /// The region the incoming request must possess to be considered for forwarding
+    pub region: String,
+
+    /// The name of the regional service to forward this request to
+    pub destinationHost: String,
```
clux: Should `destinationHost` not have validation here? Currently, the `region` property can be one of the defined "regions" based on the regex, but there's nothing stopping people from putting a full fqdn to anywhere.

clux:
most of these comments i've left now are kind of nitpicky. but 2 of them i do care about a bit, namely:

- name of the key `region` (we shouldn't overload that term)
- scope of destinationHost - it's fine if it can be a fqdn, as long as that's the intent - left another comment on the manifest pr because it feels a bit unclear

shipcat_definitions/src/manifest.rs
chris-denman-babylon: Changed to 'verify_destination_rules'.

shipcat_definitions/src/manifest.rs
chris-denman-babylon: Cool, that's cleaner.

shipcat_definitions/src/manifest.rs
chris-denman-babylon: Done

shipcat_definitions/src/structs/destinationrule.rs
chris-denman-babylon: Changed, after tribe discussion, i'll look forward to your next round of comments.

shipcat_definitions/src/structs/destinationrule.rs
chris-denman-babylon: That's the design.

shipcat_definitions/src/structs/destinationrule.rs
chris-denman-babylon: Cool, thanks for the comment. I've cleaned it up even more.

shipcat_definitions/src/manifest.rs
```diff
@@ -362,14 +362,14 @@ pub struct Manifest {
 
     /// Destination Rules
     ///
-    /// The intention here is that implementations (e.g. using Istio virtual services) will be
-    /// examine requests to determine if they satisfy this rule and if so, to forward them to
-    /// regional services as defined by 'destinationHost', otherwise they are left as-is.
+    /// The intention here is that implementations (e.g. using Istio's virtual services) will
+    /// examine requests to determine if they satisfy this rule and if so, redirect them to
+    /// alternative services as specified by 'host'.
```
clux: This is a lot better. I think it could do with a sentence about the intention of the `identifier` and how it might be used with header matching, but otherwise this is good.

PR #350, opened 2019-11-07
Fix port-forwarding kani check

jackatbancast:
There was an issue introduced where people who didn't have broader access in clusters would be barred from port-forwarding via `shipcat`. This was introduced in #327, as the check for `can-i` was incorrectly scoped.

This PR uses the `kube-rs` client directly to communicate with the cluster and properly check if you are allowed to create a `portforward` in the cluster.

Also added is a function which fetches a `Vec<ResourceRule>`, which will be used in an upcoming PR via a command line switch .


shipcat_cli/Cargo.toml
```diff
@@ -39,6 +39,7 @@ serde = "1.0.92"
 serde_derive = "1.0.92"
 serde_json = "1.0.32"
 serde_yaml = "0.8.9"
+k8s-openapi = { version = "0.6.0", features = ["v1_13"] }
```
clux: oh, you're pulling in this beast. then don't pull in default features please, it makes shipcat enormous, and kube is already an api client

shipcat_cli/src/kubectl.rs
```diff
@@ -31,18 +48,75 @@ fn kout(args: Vec<String>) -> Result<(String, bool)> {
...
+                name: None,
+                version: None,
+            }),
+            non_resource_attributes: None,
+        },
+        status: Some(SubjectAccessReviewStatus::default()),
+    };
+
+    let req = ssrr.create(&pp, serde_json::to_vec(&auth_req)?).expect("failed to create request");
+    let o = client.request::<Object<SelfSubjectAccessReviewSpec, SubjectAccessReviewStatus>>(req).map_err(ErrorKind::KubeError)?;
```
clux: oh, you're not doing the `find` anymore? I guess that makes sense. I think that method would've needed to be smarter for people with star star permissions.

shipcat_cli/src/kubectl.rs
```diff
@@ -31,18 +48,75 @@ fn kout(args: Vec<String>) -> Result<(String, bool)> {
...
+
+    let req = ssrr.create(&pp, serde_json::to_vec(&auth_req)?).expect("failed to create request");
+    let o = client.request::<Object<SelfSubjectRulesReviewSpec, SubjectRulesReviewStatus>>(req).map_err(ErrorKind::KubeError)?;
+
+    debug!("spec: {:?}", o.spec);
+    debug!("status: {:?}", o.status);
+    Ok(o.status.expect("expected rules").resource_rules)
+}
+
+fn kani(namespace: String, verb: String, resource: String, subresource: Option<String>) -> Result<bool> {
```
clux: i'd make a struct for this stuff rather than having 4 string parameters in a row. dat is not type safety.

shipcat_cli/src/kubectl.rs
jackatbancast: Since I discovered the actual reason for the issue, we don't actually need to do the `find`, but I left the function to get the permission list to use for login as we talked about before.

shipcat_cli/Cargo.toml
jackatbancast:
```suggestion
k8s-openapi = { version = "0.6.0", features = ["v1_13"], default-features = false }
```

PR #351, opened 2019-11-07
shipcat top -o yaml

clux:
for parseable output.


shipcat_cli/src/top.rs
```diff
@@ -125,29 +150,59 @@ fn sort_and_print_resources(mfs_: Vec<(Manifest, ResourceTotals)>, order: Resour
...
-                width = 8,
-            );
+    match formatting {
+        OutputFormat::Table => {
+            println!("{0:<50} {1:<8} {2:<8} {3:40}", "SERVICE", "CPU", "MEMORY", "TEAM");
+            mfs.iter().for_each(|(mf, r)| {
+                debug!("{}: cpu {} + {}, mem: {} + {}", mf.name, r.base.requests.cpu,
+                    r.extra.requests.cpu, r.base.requests.memory, r.extra.requests.memory);
+                if upper_bounds {
+                    let ub_cpu = (1000.0*(r.base.requests.cpu + r.extra.requests.cpu)) as u64;
```
skrytt: A small amount of calcs here are done for both table and YAML formatting, wonder whether refactoring them to common code could be possible, but I guess it's only a few lines so probably only if more fields get added over time

shipcat_cli/src/top.rs
```diff
@@ -125,29 +150,59 @@ fn sort_and_print_resources(mfs_: Vec<(Manifest, ResourceTotals)>, order: Resour
...
+            #[derive(Serialize)]
+            struct YamlOutput {
+                service: String,
+                team: String,
+                cpu: u64,
+                memory: u64,
+            }
+            let output = mfs.iter().map(|(mf, r)| {
+                let (cpu, memory) = if upper_bounds {
+                    let ub_cpu = (1000.0*(r.base.requests.cpu + r.extra.requests.cpu)) as u64;
```
JoseAlban: just minor: maybe extract maths like these onto smaller reusable functions? it also helps when reading, as func name would document what it's doing

clux: yep. good call. done.

shipcat_cli/src/top.rs
clux: good call. have refactored.

PR #352, opened 2019-11-11
auth bugfix + warnings + environment in templates

clux:
minor tinty things that came up today


shipcat_cli/src/kubectl.rs
```diff
@@ -13,11 +12,10 @@ use kube::{
 };
 
 use k8s_openapi::api::authorization::v1::{
-    ResourceRule,
+    //ResourceRule,
```
janslow: Can we just delete these?

clux: they are used by a commented out function - i was going to ask jack

clux: @jackatbancast

jackatbancast: Happy for it to be removed, can revert it if we end up needing it

PR #353, opened 2019-11-13
Add region's environment to template context

janslow:


clux:
duplicate of stuff in https://github.com/babylonhealth/shipcat/pull/352/files - although that covers more.
whichever you are happy with merging.

shipcat_definitions/src/template.rs
```diff
@@ -74,6 +74,7 @@ impl Manifest {
 
         ctx.insert("env", &full_env);
         ctx.insert("service", &self.name.clone());
+        ctx.insert("environment", &reg.environment.to_string().clone());
```
clux: to_string already makes a new string

PR #354, opened 2019-11-15
Remove teams support in shipcat.conf

clux:
Last teams removed internally. Drops parsing of internal teams struct.

It also ports over `shipcat get resources` into more useful `shipcat top` flags.
In particular:

```
shipcat top --squads
shipcat top --tribes
shipcat top --squads --world -oyaml
```

etc.


clux:
In the next PR we will reintroduce notificationMode as a per-service flag.

shipcat_cli/src/get.rs
```diff
@@ -51,33 +51,6 @@ pub fn codeowners(conf: &Config) -> Result<Vec<String>> {
         let md = mf.metadata;
         let mut ghids = vec![];
         match conf.serviceOwnership {
```
janslow: Can we get rid of this `match`?

shipcat_cli/src/lib.rs
```diff
@@ -82,7 +82,7 @@ error_chain! {
 
 pub use shipcat_definitions::{Manifest, ConfigType};
 pub use shipcat_definitions::structs;
-pub use shipcat_definitions::config::{self, ConfigFallback, Config, Team};
+pub use shipcat_definitions::config::{self, ConfigFallback, Config};
```
janslow: Why is this exporting anything?

clux: for main.rs - which uses shipcat_cli's lib.

shipcat_cli/src/get.rs
clux: Wasn't sure if we were going to do a chapter/tribe for owners, so left the match in, but happy to kill it.

shipcat_definitions/src/config.rs
```diff
@@ -185,10 +124,6 @@ pub struct Config {
     /// Gihub parameters
     pub github: GithubParameters,
 
-    /// Team definitions
-    #[serde(default)]
-    pub teams: Vec<Team>,
-
     /// Allowed labels
     #[serde(default)]
     pub allowedLabels: Vec<String>,
```
janslow: Can you get rid of `serviceOwnership` as well (or make it optional)?

shipcat_cli/src/get.rs
janslow: Surely that won't need a global `serviceOwnership` property though?

I guess it'd be a `chapter` or `tribe` property on the service metadata, which is mutually exclusive with `squad`

clux: We will need the enum if we want to introduce the choice, and it doesn't feel like that's decided yet. It's easy enough to remove once it is though.

shipcat_definitions/src/config.rs
clux: Ah, it is already optional because it implements default.

shipcat_cli/src/get.rs
janslow: Does it need a flag at all though? It's not a breaking change.

clux: I suppose not, it's just annoying to have to potentially re-do it if we do make a decision.

janslow: It's annoying to have `if (true) doThing()` too :P

clux: Yeh. Alright, killing it.

PR #360, opened 2019-11-25
[HANDS-49] add error optional

RenatoSerra22:
The current version of the plugin will fail requests that dont contain a user uuid in the header.
We want to make this validation optional, so we can start with it as non mandatory and as clients adjust make it mandatory.


shipcat_definitions/src/structs/kong.rs
```diff
@@ -149,6 +149,7 @@ pub struct BabylonAuthHeader {
 pub struct PiiRegionHeader {
     pub region_service_uri: String,
     pub enabled: bool,
+    pub bad_request_error: bool,
```
clux: maybe slap a #[serde(default)]` on this since otherwise it's a breaking change

RenatoSerra22:
We decided we no longer need this :)

PR #361, opened 2019-11-27
cli apply should disallow deploy to unconfigured region

clux:
disallows cli deploys to rolling environments.


clux:
although this should arguably be done at the `template` level in case we are not deploying through `shipcat apply` in the future.

PR #364, opened 2020-01-06
[DIS-314] Add support to scheme in probes HttpGet definition

fnobilia:
Upon defining a probe for a *Shipcat* worker, the default scheme used to `GET` from the specified end point is HTTPS. In case your worker does not support HTTPS, you will see kube events like
```
Warning  Unhealthy  5s (x4 over 35s)  kubelet, ip-XXX-XXX-XXX-XXX Readiness probe failed: Get https://XXX.XXX.XXX.XXX:8080/health: http: server gave HTTP response to HTTPS client
```
To fix this issue, `HttpGet` struct should support `scheme` definition as specified in the official [kubernetes doc](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes). Default value will be `HTTP` has reported in kubernetes doc.


shipcat_definitions/src/structs/probes.rs
```diff
@@ -9,11 +9,15 @@ pub struct HttpGet {
...
     pub port: String,
+    /// Scheme to use (i.e. HTTP or HTTPS)
+    #[serde(default = "http_get_default_scheme")]
+    pub scheme: String,
     /// Headers to set
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub httpHeaders: Vec<HttpHeader>,
 }
 fn http_get_default_port() -> String { "http".into() }
+fn http_get_default_scheme() -> String { "HTTP".to_string() }
```
gleaver: It might be better to have this read in as an `Option`. If we default the value here, the output of shipcat template will change for every single service defining a probe to add `scheme: HTTP`, rather than relying on kubernetes to default. 

I don't know if the k8s diff is smart enough to ignore that, but might result in some nop-reconciles.

gleaver:
Can you link the corresponding chart change PR as well?

fnobilia:
I don't think we need any change, when

PR #366, opened 2020-01-07
[DEVP-1171] Remove enum from dataHandling field names to allow looser…

gleaver:
… definitions of field names

Change the `field.name` and `process.field` fields to allow free text as the name, rather than one of a defined enum. The derivation of PII/SPII is removed, as this isn't being used.

The new validation assures that:
1) Field names are `CamelCase`, starting with a capital letter
2) Field names being used in a process are defined in a field.


shipcat_definitions/src/structs/security.rs
```diff
@@ -162,28 +108,28 @@ pub struct DataField {
...
-    pub field: DataFieldType,
+    pub field: String,
     /// Service source service for this information
     pub source: String,
 }
 
 impl DataHandling {
     pub fn verify(&self) -> Result<()> {
+        // field names must be CamelCase
+        let re = Regex::new(r"^[A-Z][[:alpha:]\d]+$").unwrap();
```
clux: Ah, good, this should avoid the empty case as well.

shipcat_definitions/src/structs/security.rs
```diff
@@ -162,28 +108,28 @@ pub struct DataField {
...
     /// Canonical field name
-    pub field: DataFieldType,
+    pub field: String,
     /// Service source service for this information
     pub source: String,
 }
 
 impl DataHandling {
     pub fn verify(&self) -> Result<()> {
+        // field names must be CamelCase
```
clux: PascalCase != camelCase

gleaver: I've never come across the term PascalCase - only UpperCamelCase and lowerCamelCase.

PR #371, opened 2020-01-14
[SIM-66] Check minReplicas parameter on autoScaling during rollout.

alexmorley:
Related to: https://github.com/babylonhealth/shipcat/issues/368


shipcat_cli/src/kubectl.rs
```diff
@@ -172,10 +172,26 @@ fn rollout_status(mf: &Manifest) -> Result<bool> {
...
         "--watch=false".into(), // always just print current status
     ];
+
+    let minReplicas;
+    if mf.autoScaling.is_some() {
+        minReplicas = mf.autoScaling.as_ref().unwrap().minReplicas;
+    } else {
+        minReplicas = mf.replicaCount.unwrap();
+    }
+    let (_, currentReplicas) = get_classified_pods(PodClassification::Active, mf).unwrap();
```
alexmorley: Are all active pods necessarily from the current rollout (e.g. are we sure that there are no old replicas - they *should* be terminating but maybe not if rolling-updates are on).

shipcat_cli/src/kubectl.rs
```diff
@@ -172,10 +172,26 @@ fn rollout_status(mf: &Manifest) -> Result<bool> {
         format!("-n={}", mf.namespace),
         "--watch=false".into(), // always just print current status
     ];
+
+    let minReplicas;
+    if mf.autoScaling.is_some() {
```
clux: you can assign this if!
`let minReplicas = if .. {} else {};`

shipcat_cli/src/kubectl.rs
clux: if there are old replicas still there, that's probably okay to ignore atm. I can't think of a case where they actually linger after a rolling update is finished at least... :thinking:

shipcat_cli/src/kubectl.rs
```diff
@@ -172,10 +172,26 @@ fn rollout_status(mf: &Manifest) -> Result<bool> {
...
+    let (_, currentReplicas) = get_classified_pods(PodClassification::Active, mf).unwrap();
+
     let (rollres, _) = kout(statusvec)?;
     debug!("{}", rollres);
     if rollres.contains("successfully rolled out") {
         Ok(true)
+    } else if currentReplicas.len() as u32 >= minReplicas {
+        debug!(
+            "Enough replicas rolled out.
+            currentReplicas {} > minReplicas {}.",
```
clux: `>=`

shipcat_cli/src/kubectl.rs
alexmorley: but this will be run during the rollout right. Just want to make sure the following situation isn't possible:
1. 2 old pods
2. 1 old pod terminating, 1 old pod running, 1 new pod running
--> shipcat thinks rollout was successful and returns as we have two running pods
3. next new pod cannot start, just 1 new pod running
--> now have a cluster with fewer than min replicas pods but rollout status was successful

alexmorley:
@clux I was thinking of writing some tests with some example kubectl output but I reckon it'll require larger changes that should be done separately (need to work out how best to mock the `kout` function). In the mean time does this need any further tests?

clux:
Yeah, this is one of those things that aren't covered too well by tests at the moment. We should test it against a real cluster before releasing it (be it minikube or a dev cluster).

alexmorley:
@clux I'll test against dev today as we're gonna be deploying to it a few times today anyway 👍

alexmorley:
Tested on minikube. There are a couple of cases where it gives successful deploy even where there is a failure. Will fix.

shipcat_cli/src/kubectl.rs
clux: Oh.. that is actually possible, crap.

You could limit the pod search to the latest replicaset (e.g. `hippo-kb-graph-7cfcc4559c-p57vv` has replicaset hash == `7cfcc4559c` and pod hash == `p57vv`) - though that is not safe against multiple simultaneous upgrades (last one wins). Might have to search for the specific replicaset that was created by our `apply` call (though this leads to a possible new error case where one upgrade is Preceeded by another upgrade while this was in progress).

clux: It might be easiest to have look for at least `N` pods with the same replicaset hash.

The only case this doesn't fully catch is when there are multiple upgrades happening at the same time and then you either have:

- the other upgrade wins slowly (first upgrade waiters will report failure, second will succeed)
- the other upgrade wins quickly (both upgrade waiters see the last success simultaneously)

clux:
Did you give up on this? I have a bunch of related things to look at soon, but would still appreciate this PR.

alexmorley:
> Did you give up on this? I have a bunch of related things to look at soon, but would still appreciate this PR.
I have and hour and a half blocked out to look at it again tomorrow morning!

clux:
Replaced in #428

PR #372, opened 2020-01-16
ARCH-1: Add context to manifest.

paulseddon:


shipcat_filebacked/src/manifest.rs
```diff
@@ -220,11 +222,13 @@ impl ManifestSource {
         // TODO: Remove and use folder name
         let name = self.name.clone().require("name")?;
         let metadata = self.build_metadata(conf)?;
+        let context = self.context.clone().unwrap_or("unassigned".to_string());
```
clux: because you need to create a default ("unassigned"), here, I would take this out of BaseManifest (which is an internal type anyway)

paulseddon: Done.

shipcat_filebacked/src/manifest.rs
```diff
@@ -211,6 +213,7 @@ impl ManifestSource {
             // TODO: Make image non-optional
             image: Some(self.build_image(&base.name)?),
             version: overrides.version.build(&())?,
+            context: self.context.clone().unwrap_or("unassigned".to_string()),
```
clux: this should just be a .clone() rather than trying to force it into a magic string

shipcat_definitions/src/manifest.rs
```diff
@@ -111,6 +111,18 @@ pub struct Manifest {
...
+    /// The bounded context that this service belongs to
+    ///
+    /// Informational use only, referenced by external tooling in order
+    /// to create a software map of domains and constituent services.
+    ///
+    ///
+    /// ```yaml
+    /// context: consultation
+    /// ```
+    #[serde(default)]
```
clux: you probably also want a `skip_serializing_if " Option::is_none"` if it's wrapped in an Option. That way you either get something in the CRD, or nothing => No more CRDs change than the ones you actually applied the property to.

paulseddon: I'll need to update the Deployment base chart then to write the `context` if it exists. Will do separately.

paulseddon: Actually I only really need it on the CRD for now, which it is.

clux:
Fixed your merge conflict (again), please pull next time :-)

shipcat_definitions/src/manifest.rs
clux: Ok, good :-)

PR #380, opened 2020-01-24
manifest.workload + statefulset basic support

clux:
**EDIT: IGNORE THIS COMMENT READ COMMENTS BELOW**
** WE ARE NO LONGER TRYING TO SHOEHORN VOLUMES INTO DEPLOYMENTS **

there's a dependent pr for this in our manifests. Required to make this work is something like this in the `deployment.yml` part of the chart:

first, under deployment's `volumeMounts`:
```yaml
      {{- range $p := $.Values.persistentVolumes }}
        - name: {{ $p.name }}
          mountPath: {{ $p.mountPath }}
      {{- end }}
```

secondly, making sure it exists, under deployment's `volumes`:

```yaml
      {{- range $p := $.Values.persistentVolumes }}
      - name: {{ $p.name }}
        persistentVolumeClaim:
          claimName: {{ $p.name }}
      {{- end }}
```

and weirdly, under deployments `initContainers`:

```yaml
{{- range $p := $.Values.persistentVolumes }}
          initContainers:
          - name: {{ $p.name }}-mount-permissions
            image: busybox
            command: ["sh", "-c", "chmod 777 {{ $p.mountPath }}"]
            volumeMounts:
            - name: {{ $p.name }}
              mountPath: {{ $p.mountPath }}
{{- 
```

This part is because the mounted volume is root-owned by default, which breaks for secure non-root containers. See https://serverfault.com/questions/906083/how-to-mount-volume-with-specific-uid-in-kubernetes-pod

And finally, the resource that creates it, either stand-alone, or at the bottom of `deployment.yml`:

```yaml
{{- range $p := $.Values.persistentVolumes }}
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: {{ $p.name }}
  namespace: {{ $.Values.namespace }}
  labels:
    app: {{ $.Values.name }}
    {{- template "chart.squadAndTribeLabelsIndent4" $ }}
{{- template "chart.ownerRefs" $ }}
spec:
  accessModes:
    - {{ $p.accessMode }}
  storageClassName: gp2
  resources:
    requests:
      storage: {{ $p.size }}
{{- end }}
```


shipcat_definitions/src/structs/resources.rs
```diff
@@ -151,12 +151,22 @@ pub fn parse_memory(s: &str) -> Result<f64> {
         res *= 1024.0*1024.0;
     } else if unit == "Gi" {
         res *= 1024.0*1024.0*1024.0;
+    } else if unit == "Ti" {
+        res *= 1024.0*1024.0*1024.0*1024.0;
+    } else if unit == "Ti" {
```
gleaver: copy pasty

shipcat_definitions/src/structs/resources.rs
```diff
@@ -151,12 +151,22 @@ pub fn parse_memory(s: &str) -> Result<f64> {
         res *= 1024.0*1024.0;
     } else if unit == "Gi" {
         res *= 1024.0*1024.0*1024.0;
+    } else if unit == "Ti" {
```
gleaver: is `Ki` correct?

clux: Ugh, from kube docs:
> You can express memory as a plain integer or as a fixed-point integer using one of these suffixes: E, P, T, G, M, K. You can also use the power-of-two equivalents: Ei, Pi, Ti, Gi, Mi, Ki.

so it doesn't even follow SI. Uppercase both K and Ki.

clux:
Actually. Going to redo this so we can properly support StatefulSet instead. It's not very useful to have have Pvcs in the base chart if we can't use rolling upgrade with the default RWO access mode for EBS.

clux:
Considering fully adding the concept of a `kind` to a shipcatmanifest. This allows the charts to do intelligent multiplexing and shipcat to share common concepts between charts and and internal logic. See https://github.com/babylonhealth/shipcat/issues/382, will probably proceed with this unless there are no objections.

clux:
To avoid the confusion of having a `kind` inside a `spec` from a k8s perpective, we have renamed the enum to the more actual `workload` to map to a `PrimaryWorkload` enum.

PR #388, opened 2020-02-06
Extend audit+webhooks with a Delete Event

### LINKED ISSUE #376: undeploying services should notify
Currently removing an entry in the `manifest.regions` vector will cause an undeploy, but no notification. Add one to the part of cluster.rs that does the undeployment after the hashset compare.

clux:
Fixes #376 

In the process, this is also majorly cleaning up the audit.rs module, which
- was awkwardly spread out between tests folder (just to load a manifest we didn't care about), and the tiny impl
- had a pointless trait that should have been an enum
- had functions that should have been impls on the main Audit type
- had test functionality commented out (reinstated with new version of mockito)


shipcat_cli/src/status.rs
```diff
@@ -93,6 +96,14 @@ impl Status {
         Ok(o)
     }
 
+    /// Minimil CRD deleter
```
bh-anton-k:
```suggestion
    /// Minimal CRD deleter
```

Issue #395, opened 2020-02-12
Replace upgrade checks with kube api calls


clux:
We currently badly scrape `kubectl describe deployment` output for Events and Conditions using regexes, kind of like users do with `grep`. We should instead do this with the `kube` api library, and check the .status objects for the actual data instead.

Marking this as an issue now because we have other work to make this work better anyway. I don't want us to further extend the shell scraping logic.

However, this requires us upgrading kube to 0.25.0, which requires us upgrading to async #394 first.

Issue #394, opened 2020-02-12
Upgrade main to async main


clux:
So we can have reqwest 0.10 etc.

We should __try__ to do this without upgrading anything but the executor. It will be hairy if we try to rewrite everything in the same PR.

I think we should not to convert all the FILE access stuff to async right away, or if we do, just use block_on inside the parallelised cluster module. See #384 for an idea of what's needed. This is going to be the hardest part because it currently uses `rayon`.

We should try to do it with tokio stream features rather than what's in async_std atm. Once those two are done, upgrading reqwest should be a lot easier.

clux:
Done in #401

PR #401, opened 2020-02-17
Move to async with tokio 0.2

clux:
This is to move to the async ecosystem to get:

- reqwest 0.10
- kube (like the last 8 versions)

and to be able to rely on async stream mechanics for parallelism rather than forcing rayon to do it badly.

This is actually performing really well so far for `shipcat cluster diff` and `shipcat cluster crd reconcile`, so might just have this as a start. The rest can be asyncified at will.


clux:
replaces #384 because async_std was more awkward to deal with.

fixes #395 and #394

clux:
Going to leave this as is without trying to unify the reqwest versions (which will reduce the dependency tree quite a bit), because the PR is getting quite big.

shipcat_cli/src/top.rs
```diff
@@ -38,51 +38,66 @@ impl FromStr for ResourceOrder {
...
-            }
-            if let Some(mf) = first_mf {
-                Ok(Some((mf, res)))
-            } else {
-                Ok(None)
+async fn calculate_manifest_requests(conf: &Config, reg: &Region) -> Result<Vec<(Manifest, ResourceTotals)>> {
+    let available = shipcat_filebacked::available(conf, &reg).await?;
+    let mut buffered = stream::iter(available)
+        .map(move |mf| load_mf_req(mf.base.name, conf, reg))
+        .buffer_unordered(100);
```
bh-anton-k: 100? 🤔I though it should be bound by core count, practical value related to concurrent TCP connections/in-flight requests we expect to be keeping to kube backend or something, I dunno, looks like a magick number to me

clux: Yeah, I am fixing this in the next pr

PR #402, opened 2020-02-18
upgrade reqwest to 0.10

clux:
two parts:

- in shipcat_cli and shipcat_definitions for internal http calls
- via a version bump to self_update


clux:
Replaced in #403

PR #403, opened 2020-02-18
upgrade reqwest to 0.10

clux:
- [x] shipcat_cli self_update to 0.13 for reqwest 0.10
- [x] shipcat_status reqwest to 0.10 for vault.rs
- [x] shipcat_cli reqwest to 0.10 for audit.rs
- [x] slack_hook upgrade for reqwest 0.10
- [x] raftcat reqwest to 0.10 for integrations.rs

ends up removing like 30 deps from the cargo tree of shipcat_cli


clux:
Hm. Needs a little work on self_update still. This has broken it :/

clux:
Trying to work around https://github.com/jaemk/self_update/issues/44

clux:
Alright, have replaced the self_update dependency. It was only doing 2 http requests + tarball extraction in our case anyway. Added about a 100 lines to upgrade.rs, but a lot of that was because I was trying to be fancy with indicatif (plan on using that for upgrade status anyway).

raftcat/Cargo.toml
```diff
@@ -15,24 +15,26 @@ path = "src/lib.rs"
...
+kube = "0.25.0"
 #kube = { path = "../../kube-rs" }
 #kube = { git = "https://github.com/clux/kube-rs", rev = "14a851d115ed8ea8ddeeb46398450d25862e3c11" }
 serde_json = "1.0.32"
 serde_yaml = "0.8.9"
 serde = "1.0.92"
 serde_derive = "1.0.92"
-prometheus = "0.4"
-url = "1.7.2"
+#prometheus = "0.4"
```
janslow: Can these commented deps just be removed?

clux: I would like to re-introduce sentry again, but the integration is broken in actix >= 1 (have a bug open for it). but guess it doesn't need to be there. removed.

PR #406, opened 2020-02-21
rip out openssl for rustls - fixes #329

### LINKED ISSUE #329: swap over from openssl to rustls
For various reasons:

- version pinning on macos can break autoupgrade
- rustls can compile statically without my hacky build container
- openssl is shit

clux:


clux:
seems refreshing the Cargo.lock file caused us to get a new format, so everything looks like it's changed. dependency tree is smaller still!

.circleci/config.yml
```diff
@@ -80,10 +79,10 @@ jobs:
       - RUST_BACKTRACE: "full"
     steps:
       - checkout
-      - run: grep "\[metadata\]" -A 9000 Cargo.lock > deps_checksum
+      - run: make sanity
```
janslow: 😆

PR #407, opened 2020-02-21
Fix upgrade flate errors

clux:
Occasionally we got gzip errors if not doing this. Not fully been able to reproduce why... Maybe it's due to how we made the archive:

```sh
tar czf shipcat.${ver}.tar.gz --transform=s,^${ver}/,, $(find ${ver}/ -type f -o -type l)
```


shipcat_cli/src/upgrade.rs
```diff
@@ -246,7 +248,14 @@ fn extract_tarball<T: AsRef<Path>>(extract_path: &Path, src: T) -> Result<()> {
...
-    archive.unpack(&extract_path)?;
 
+    let mut entry = archive
+        .entries()?
+        .filter_map(|e| e.ok())
+        .find(|e| e.path().ok().filter(|p| p == file_to_extract).is_some())
+        .ok_or_else(|| {
+            ErrorKind::SelfUpgradeError(format!("Could not find path in archive: {:?}", file_to_extract))
+        })?;
+    entry.unpack_in(&extract_path)?;
```
clux: this is the only real change, going back to what self_update did, and only extracting the main bin from the archive

shipcat_cli/src/upgrade.rs
```diff
@@ -130,7 +130,8 @@ pub async fn self_upgrade(ver: Option<Version>) -> Result<()> {
     // Because upgrade fs::rename call can fail when moving across partitions
     // we try to avoid this by using a temp dir inside the path shipcat is found..
     let tmp_dir = TempDir::new_in(exe.base, "shipcat_upgrade")?;
-    std::fs::create_dir_all(tmp_dir.path())?;
+    //let tmp_dir = exe.base.join(&release.tag_name);
```
janslow: Do you need this line?

clux: removed

PR #417, opened 2020-02-27
eventdefintions a manditory parameter

nandaloussiBabylon:


shipcat_definitions/src/structs/eventstream.rs
```diff
@@ -18,8 +18,7 @@ pub struct EventStream {
     #[serde(default)]
     pub consumers: Vec<String>,
 
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub event_definitions: Option<Vec<EventDefinition>>,
+    pub event_definitions: Vec<EventDefinition>,
```
clux: ok, not even a default if it's not set?
people can still explicitly set it to the empty vector.

PR #418, opened 2020-02-27
make cluster check usable with --skip-kinds

clux:


shipcat_cli/src/main.rs
```diff
@@ -262,6 +264,11 @@ fn build_cli() -> App<'static, 'static> {
                 .short("c")
                 .long("check")
                 .help("Check the validity of the template"))
+               .arg(Arg::with_name("skip-kinds")
+                .long("skip-kinds")
+                .takes_value(true)
+                .requires("check")
+                .help("Kinds to ignore strongest checks for for (comma separated)"))
```
gleaver: 4 4

shipcat_cli/src/helm.rs
```diff
@@ -99,7 +99,7 @@ pub async fn template(mf: &Manifest, output: Option<PathBuf>) -> Result<String>
 /// We don't validate kubernetes schemas in here, but we do validate consistency of:
 /// - labels: app.kubernetes.io/name, app.kubernetes.io/version, app.kubernetes.io/managed-by
 /// - ownerReferences (need ShipcatManifest, !controller, uid propagated, name correct)
-pub fn template_check(mf: &Manifest, reg: &Region, tpl: &str) -> Result<()> {
+pub fn template_check(mf: &Manifest, reg: &Region, skipped: &Vec<String>, tpl: &str) -> Result<()> {
```
gleaver: is there a reason this one isn't async?

shipcat_cli/src/helm.rs
```diff
@@ -184,7 +184,7 @@ fn check_labels(mf: &Manifest, kind: &str, obj: &MetaObject) -> Result<bool> {
     // If the object doesn't get injected into the Deployment automatically
     // then it ought to have the standard version property.
     // If it changes, we should not lie about it changing (Secret + CM didn't really change)
-    if !["Secret", "ConfigMap"].contains(&kind) {
+    if !["Secret", "ConfigMap"].contains(&kind) && !skipped.contains(&kind.to_string()) {
```
gleaver: If we support `skip-kinds`, should these still be hardcoded?

shipcat_cli/src/cluster.rs
```diff
@@ -97,19 +97,18 @@ async fn check_summary(svc: String, conf: &Config, reg: &Region) -> Result<Strin
...
 ///
 /// Helper that shells out to helm template in parallel.
-pub async fn mass_template_verify(conf: &Config, reg: &Region) -> Result<()> {
+pub async fn mass_template_verify(conf: &Config, reg: &Region, skipped: &Vec<String>) -> Result<()> {
     let svcs = shipcat_filebacked::available(conf, reg).await?;
-    assert!(conf.has_secrets());
 
     let mut buffered = stream::iter(svcs)
-        .map(move |mf| check_summary(mf.base.name, &conf, &reg))
+        .map(move |mf| check_summary(mf.base.name, &skipped, &conf, &reg))
```
gleaver: doesn't this need `await`ing?

shipcat_cli/src/helm.rs
clux: yep, is has all data and it's not doing IO

shipcat_cli/src/cluster.rs
clux: No :-)

clux: No :-)

Missing await is only a problem if we explicitly trashed the output var  (like by doing `let _ = fncall()`)

shipcat_cli/src/helm.rs
clux: Probably. It's kind of assumed that deployments change when secrets/configs change.

PR #428, opened 2020-03-18
Fix upgrade tracking for hpas

### LINKED ISSUE #368: when using autoscaling, shipcat might think rollout has failed when it hasnt
Becacuse the rollout might trigger autoscaling conditions, there might be out-of-band rollouts happening while a standard rolling upgrade is happening. This might result in the deployment not having 100% ready pods after the wait period, and thus shipcat will call the upgrade a failure, and send notifications.

Instead, we should check that we have at least N ready pods at each wait point (where `N` is `autoScaling.minReplicas.unwrap_or(replicaCount)`).

Should be fairly easy to do, we just need to pass those two manifest parameters (if &Manifest isn't already passed) to the function that greps the kubectl get pods output during upgrades).

### LINKED ISSUE #379: shipcat base deployment chart handles replicacount badly with autoscaling
We get tons of reports of replicacount changing when using an HPA via the autoScaling syntax.

We should probably only set deployment replicas if we don't have autoscaling set. Try it out.

### LINKED ISSUE #395: Replace upgrade checks with kube api calls

clux:
Fixes #379 and fixes #368 and fixes #395.

This is done through a set of steps:

- [x] upgrades kube to 0.30.0 so we can use the api to get better info
- [x] replaces a bunch of tracking calls from `kubectl` to the `kubeapi` module
- [x] adds a `shipcat cluster crd install` because `kube` can generate crds
- [x] remove almost all of the crds.rs logic
- [x] makes a `kubeapi` module from the shell of `status`
- [x] replaces `kpods` with a better status tracker using the api (WIP during porting)
- [x] replaces upgrade trackers with a smart variant

The smart upgrade tracker variant reads a deployment and replicaset at similar intervals as when we did kubectl get and (and then did regex parsing of it). Now we track the actual last created replicaset (so we can even see if an upgrade got succeeded (we are not really catching that atm though)), and create a progress bar of how far it is along.

Outstanding:
- [x] test failures
- [x] raftcat build
- [x] track conditions on deployment during rollout (regression)


clux:
Hm. Ok, one bug still here. 
> thread 'main' panicked at 'attempt to subtract with overflow', shipcat_definitions/src/structs/rollingupdate.rs:126:13

clux:
conditions is actually being shown on failure anyway:

```
shipcat::track: Fetching logs from non-ready main container in pod: raftcat-79fcbb7d89-bqjth
shipcat::track: Failed to get logs from raftcat-79fcbb7d89-bqjth: kube api: ApiError: container "raftcat" in pod "raftcat-79fcbb7d89-bqjth" is waiting to start: trying and failing to pull image: BadRequest (ErrorResponse { status: "Failure", message: "container \"raftcat\" in pod \"raftcat-79fcbb7d89-bqjth\" is waiting to start: trying and failing to pull image", reason: "BadRequest", code: 400 }): Api(ErrorResponse { status: "Failure", message: "container \"raftcat\" in pod \"raftcat-79fcbb7d89-bqjth\" is waiting to start: trying and failing to pull image", reason: "BadRequest", code: 400 })
```

clux:
This replaces the viewer logic we relied on from [kubectl rollout](https://github.com/kubernetes/kubectl/blob/d134c5a98730a019115f44dadb8d08679ce51641/pkg/polymorphichelpers/rollout_status.go#L76-L89).

Issue #437, opened 2020-03-24
api gc can fail to deserialize status output


clux:
Not sure what all the preconditions for this is, but have seen cluster.rs attempt to garbage collect disabled services and fail:

```sh
17:23:55 shipcat::cluster: Will remove excess manifests: ["myservice"]
17:23:55 kube::client: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Success","details":{"name":"myservice","group":"babylontech.co.uk","kind":"shipcatmanifests","uid":"a66a7c7f-6dca-11ea-8d88-0a71ead2e0e0"}}
17:23:55 , Error("missing field `name`", line: 1, column: 204)
```

It's probably from the `kube` library upgrade and it's probably too strict with the `Status` struct.

clux:
hm, it `name` is optional:

https://github.com/clux/kube-rs/blob/bc0153a16bc9617e6ff875d862f2423300994cb4/kube/src/client/mod.rs#L15

clux:
Yep, definitely not it: https://github.com/clux/kube-rs/commit/22c13c34bd56a4323eab0e5b923e7cb9b03e54ef

Waiting to see if it can be reproduced.

clux:
Seen again on a temporary chart I was goofing around with yesterday:

```
15:55:57 shipcat::cluster: Will remove excess manifests: ["logstash"]
15:55:58 kube::client: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Success","details":{"name":"logstash","group":"babylontech.co.uk","kind":"shipcatmanifests","uid":"a43693ac-75ba-11ea-b122-06c8557584da"}}
15:55:58 , Error("missing field `name`", line: 1, column: 197)
15:55:58 shipcat: cluster error: kube api: Error deserializing response: SerdeError(Error("missing field `name`", line: 1, column: 197))
```

did not trigger my error handling :thinking:

clux:
Problem was using the raw `Resource` api in kube badly. We asked to serialize a delete response as a manifest, when it can return `Either<Manifest, Status>` so it would crash when we received a status. This is documented better in the kube library now.

Fixed in 0.152.0

PR #444, opened 2020-04-04
[OB-393] fix delete deserialize failure #437

clux:
resurrecting branch because #437 happened again


shipcat_cli/src/kubeapi.rs
```diff
@@ -100,7 +100,7 @@ impl ShipKube {
         let dp = DeleteParams::default();
         let req = self.mfs.delete(&self.name, &dp).map_err(ErrorKind::KubeError)?;
         self.client
-            .request::<MinimalManifest>(req)
+            .request_status::<MinimalManifest>(req)
```
clux, line 103: This was the cause of the bug. Serializing into the wrong type. `request_status` can return an `Either<TYPEWESAID, Status>` so it would crash when we got `Status`

PR #448, opened 2020-04-06
[issue-447] Add statefulset rollout status revision check

nitinpatil1992:
[Issue-447](https://github.com/babylonhealth/shipcat/issues/447)
[Rust-Statefulset-KubeOpenAPI-Reference](https://docs.rs/k8s-openapi/0.7.1/k8s_openapi/api/apps/v1/struct.StatefulSetStatus.html#structfield.update_revision)


shipcat_cli/src/track.rs
```diff
@@ -282,8 +284,10 @@ impl TryFrom<StatefulSet> for StatefulSummary {
         if let Some(status) = d.status {
             let ready = status.ready_replicas.unwrap_or(0);
             let replicas = status.replicas;
+            let current_revision = status.current_revision;
+            let updated_revision = status.updated_revision;
```
htsili, line 288: it should be status.update_revision
https://docs.rs/k8s-openapi/0.7.1/k8s_openapi/api/apps/v1/struct.StatefulSetStatus.html#structfield.update_revision

shipcat_cli/src/track.rs
```diff
@@ -364,7 +368,7 @@ async fn rollout_status(mf: &Manifest, kube: &ShipKube, hash: &Option<String>) -
             let ss = kube.get_statefulset().await?;
             let s = StatefulSummary::try_from(ss)?;
             let minimum = mf.min_replicas();
-            let ok = s.ready == s.replicas;
+            let ok = s.ready == s.replicas && s.current_revision == s.updated_revision;
```
clux, line 371: it's `update_revision` we want see https://arnavion.github.io/k8s-openapi/v0.7.x/k8s_openapi/api/apps/v1/struct.StatefulSetStatus.html

and we only want to compare if they both `is_some()`.

shipcat_cli/src/track.rs
```diff
@@ -282,8 +284,10 @@ impl TryFrom<StatefulSet> for StatefulSummary {
         if let Some(status) = d.status {
             let ready = status.ready_replicas.unwrap_or(0);
             let replicas = status.replicas;
+            let current_revision = status.current_revision;
```
clux, line 287: need to get current_revision outside of this function around 406 where we get the hash for deployments. We can do that in the same `match`

shipcat_cli/src/track.rs
nitinpatil1992: Updated!

PR #452, opened 2020-04-07
[BPS-2000] always obfuscate secret output - fixes #438 fixes #451

### LINKED ISSUE #438: diff masking of secret could be better
We currently mask all secrets found in vault by a literal replace on the diff.

This works, but has 2 negative consequences:

- it reveals people using trivial passwords matching service names (by masking in an expected place)
- it reveals old passwords on changes

The latter is obviously more problematic as people use password generators.

But there is something we could do to improve this:

- [ ] parse the diff output as the corresponding kubernetes objects
- [ ] any diff to the secret objects, replace contents with "+++Secret diff concealed"

This would be analogous to what helm diff does.

### LINKED ISSUE #451: diff on secretFiles can fail to mask
related to #438 but actually not masking because of b64 obfuscation that's using a different obfuscator...

This can cause secrets to leak in logs.

We should actually just solve both these issues and filter out any secret object diff regardless of values.

clux:
New diff output from reconcile now:

```diff
extensions.v1beta1.Deployment.dev.raftcat has changed:
-        checksum/secrets: 3e6c8c2098228c15554026bef845a1e121c6d32a315ecbba953d92e8b4d26bed
+        checksum/secrets: fd7f339fc814c7067c66213705ff7bfc78a81723e50c1083bd3a29df272d34e7
Change to v1.Secret.dev.raftcat elided for security
```


shipcat_cli/src/diff.rs
```diff
@@ -410,4 +394,49 @@ v1.Service.dev has changed:
...
+diff -u -N /tmp/LIVE-422759316/v1.Secret.dev.raftcat-secrets /tmp/MERGED-101659107/v1.Secret.dev.raftcat-secrets
+--- /tmp/LIVE-422759316/v1.Secret.dev.raftcat-secrets   2020-04-07 12:26:32.618020569 +0100
++++ /tmp/MERGED-101659107/v1.Secret.dev.raftcat-secrets 2020-04-07 12:26:32.664686669 +0100
+@@ -1,9 +1,10 @@
+ apiVersion: v1
+ data:
+
+-  SENTRY_DSN: aGVsbG8gd29ybGQK==
++  SENTRY_DSN: YUdWc2JHOGdkMjl5YkdRPQ==
++  WOOT: aGk=
```
clux, line 429: fake values btw

PR #464, opened 2020-05-22
DAT-320 fix topic and user name regex

nandaloussiBabylon:


shipcat_definitions/src/structs/kafkaresources.rs
```diff
@@ -126,10 +127,26 @@ impl KafkaResources {
...
+        }
+
+        if failed_topics.is_empty() && failed_users.is_empty() {
             Ok(())
+        } else if !failed_topics.is_empty() {
+            bail!(
+                "invalid topic name(s), must match expression \"^[0-9a-z\\-\\.]{{1,63}}$\": {:?}",
+                failed_topics
+            );
+        } else if !failed_users.is_empty() {
```
clux, line 143: this does not need to be an else if, bail returns early

shipcat_definitions/src/structs/kafkaresources.rs
```diff
@@ -126,10 +127,26 @@ impl KafkaResources {
...
+            bail!(
+                "invalid user name(s), must match expression \"^[0-9a-z\\-\\.]{{1,63}}$\": {:?}",
+                failed_users
+            );
         } else {
-            bail!("invalid topic name(s): {:?}", failed_topics);
+            Ok(())
```
clux, line 149: this should be at the bottom of the fn if you use early returns

PR #467, opened 2020-05-29
OB-512 Shipcat support for Prometheus alerts

skrytt:
- Debug build compiles successfully for shipcat_definitions, shipcat_filebacked, shipcat
- Tested manually using `shipcat values` to generate values based on modified versions of the pa-orchestrator manifest for dev-uk region, including the correct usage case and various validate rule breakages
- Used shipcat template to produce Helm chart data from which I extracted just the PrometheusRule created from my testing, applied manually to dev-uk to verify this works as intended, rule is correctly picked up by Prometheus


shipcat_definitions/src/structs/prometheusalert.rs
```diff
@@ -0,0 +1,42 @@
...
+use regex::Regex;
+
+/// Data desribing one Prometheus alert.
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct PrometheusAlert {
+    name: String,
+    summary: String,
+    description: String,
+    expr: String,
+    for_: String,
```
skrytt, line 11: I was thinking about renaming `for_` to something like `min_duration`, thoughts?

mgust: Yeah, I was wondering what "for_" would be used for.

skrytt: In a Prometheus alert `for` is the name of the field for minimum duration the condition is true before an alert fires; but unfortunately it's a keyword in Rust
Going ahead and using `min_duration` instead, the manifests Chart template will use the value of `min_duration` to set `for` in the K8s PrometheusRule

shipcat_definitions/src/structs/prometheusalert.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// Data desribing one Prometheus alert.
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct PrometheusAlert {
+    name: String,
+    summary: String,
+    description: String,
+    expr: String,
+    min_duration: String,
+    severity: String,
```
clux, line 12: This should be an enum { Warning, Error } and have it use serde(rename_all(lowercase))

shipcat_definitions/src/structs/prometheusalert.rs
```diff
@@ -0,0 +1,42 @@
...
+            bail!("Prometheus alert needs an expr (PromQL) to evaluate");
+        }
+        if !Regex::new(r"^\d+[mh]$").unwrap().is_match(&self.min_duration) {
+            bail!("Prometheus alert has invalid min_duration value (needs to be like '15m' or '1h')");
+        }
+        match self.severity.as_str() {
+            "warning" => (),
+            "error" => (),
+            _ => bail!("Prometheus alert severity must be 'warning' or 'error'"),
+        }
```
clux, line 38: This falls away if you have an enum

shipcat_definitions/src/structs/prometheusalert.rs
```diff
@@ -0,0 +1,42 @@
+use super::Result;
+use regex::Regex;
+
+/// Data desribing one Prometheus alert.
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct PrometheusAlert {
+    name: String,
+    summary: String,
+    description: String,
```
clux, line 9: Docstrings here would be good. I see you have put them in the error message, but it would be good to put what they are for here with triple slashes for API docs.

shipcat_definitions/src/structs/prometheusalert.rs
```diff
@@ -1,15 +1,39 @@
...
+
+    /// The minimum duration a pending alert must remain active for in order to actually fire.
+    /// Examples: '15m', '1h'.
     min_duration: String,
-    severity: String,
+
+    /// The severity of the alert, corresponding to how urgently it should be actioned if it were
+    /// in production.
+    /// Error => High urgency, wakes up on-call engineers and escalates if unacknowledged
+    /// Warning => low urgency, does not wake on-call engineers and does not escalate.
```
clux, line 28: docstrings need a blank line after the heading, e.g.:

```
/// Severity of the alert
///
/// Corresponding to...
```

shipcat_definitions/src/structs/prometheusalert.rs
```diff
@@ -1,15 +1,39 @@
...
+    severity: PrometheusAlertSeverity,
+}
+
+#[serde(rename_all = "lowercase")] 
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub enum PrometheusAlertSeverity {
+    Warning,
+    Error,
```
clux, line 36: You can also put the docs for what Error means here, rather than in the documentation for the severity field.

PR #471, opened 2020-06-29
Statuscake owner tagging

clux:
So we can route test failures to the correct pagerduty rule in the event stack.


shipcat_cli/src/statuscake.rs
```diff
@@ -34,26 +35,30 @@ impl StatuscakeTest {
         };
 
         // Generate tags, both regional and environment
-        let mut test_tags = format!("{},{}", region.name, region.environment.to_string());
+        let mut tags = vec![];
+        tags.push(region.name.clone());
+        tags.push(region.environment.to_string());
+        tags.push(format!("squad={}", md.squad.as_ref().expect("squad exists")));
+        tags.push(format!("tribe={}", md.tribe.as_ref().expect("tribe exists")));
```
skrytt, line 42: Should the expect panic message be "squad doesn't exist", same for tribe? (it looks like it should fail if we can't get a reference to the field from the metadata, maybe I misunderstood)

clux: Yeah, you've understood it correctly. This would cause a panic like:

```
thread 'main' panicked at 'squad exists'', /home/clux/babylon/shipcat/shipcat_cli/src/statuscake.rs:67:5
```

i see the positive expectation more frequently in rust assertions like this.

We have validated merge on manifests that squad/tribes existing already, so not bothering to make better error handling for this at the moment.

skrytt: Sure thing, thanks for the explanation! :)

PR #475, opened 2020-08-05
[DXKP-120] Add kompass-hub plugin

carlosgaldino:
This patch adds a new endpoint to Raftcat `/raftcat/kompass-hub` which will be used by the hub to fetch information about the services that Raftcat is monitoring.

The interaction with `kompass-hub` is as follows:

1) Raftcat registers with the hub informing which endpoint it is exposing.
2) Hub contacts Raftcat periodically to pull information.
3) Raftcat renews the registration in step 1 by periodically sending heartbeats to the hub.

Points to note:

I didn't add the `build.rs` file used to generate the files inside `protos/` because the docker image used to build the project doesn't have `protoc` available. If it's possible to make `protoc` available then we could include the `build.rs` in the project, although it shouldn't require changes in the definition again.

The case where we need to re-register (with exponential backoff) in case of failed communication with the hub will be added in the future.


raftcat/src/kompass.rs
```diff
@@ -0,0 +1,90 @@
...
+    if auth_token.is_none() {
+        let err = format_err!("{} not set", KOMPASS_AUTH_TOKEN);
+        warn!("{}", err);
+        return Err(err);
+    }
+
+    let region = env::var("REGION_NAME").expect("Need REGION_NAME evar");
+    let ns = env::var("NAMESPACE").expect("Need NAMESPACE evar");
+    let id = format!("raftcat-{}-{}", region, ns);
+    let uri = "http://localhost:8080/raftcat/kompass-hub";
```
carlosgaldino, line 48: What is the best way of obtaining the current address of the running Raftcat instance?

raftcat/src/main.rs
```diff
@@ -254,6 +268,10 @@ async fn main() -> std::io::Result<()> {
     };
     let shared_state = state::init(cfg).await.unwrap();
 
+    tokio::spawn(kompass::register(
+        env::var("KOMPASS_URL").expect("Need KOMPASS_URL evar"),
+    ));
```
clux, line 273: parse the url here, then spawn the register loop since we never deal with it.
ideally also grab the future from register and use tokio::select on that along with the web server future:
https://github.com/clux/controller-rs/blob/master/src/main.rs#L52-L55

raftcat/Cargo.toml
```diff
@@ -13,6 +13,14 @@ path = "src/main.rs"
 name = "raftcat"
 path = "src/lib.rs"
 
+[build-dependencies]
+protoc-rust = "2.0"
+
+[features]
+default = ["with-serde"]
+
+with-serde = []
```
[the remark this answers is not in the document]
carlosgaldino: We need to activate the `with-serde` feature in order for the proto structs be serialisable: https://github.com/babylonhealth/shipcat/pull/475/files?file-filters%5B%5D=.rs&file-filters%5B%5D=.toml#diff-50fe544165fecf730601276ede361f02R27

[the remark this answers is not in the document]
clux: That makes sense, but that doesn't explain why you need to re-export `with-serde` as a feature for `raftcat`?

raftcat/src/main.rs
```diff
@@ -254,6 +273,15 @@ async fn main() -> std::io::Result<()> {
     };
     let shared_state = state::init(cfg).await.unwrap();
 
+    let region = shared_state.get_region().await.unwrap();
+    let region_str = region.raftcat_url().expect("raftcat url");
+    let region_url = Url::parse(&region_str)
+        .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, "invalid raftcat url"))?;
+    let kompass_evar = env::var("KOMPASS_URL").expect("Need KOMPASS_URL evar");
+    let kompass_url = Url::parse(&kompass_evar)
+        .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, "invalid kompass url"))?;
```
clux, line 282: Yeh, this is a reasonable error type, thanks!

raftcat/Cargo.toml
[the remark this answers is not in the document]
carlosgaldino: Isn't that the definition of a feature which is also used as a default one? Not a re-export?!

carlosgaldino:
> left some suggested tweaks. but generally looks good 👍
> 
> also a question, how is the protofbuf files generated? i don't see a source.

Because of this (in the description):

> Points to note:
>
> I didn't add the `build.rs` file used to generate the files inside `protos/` because the docker image used to build the project doesn't have `protoc` available. If it's possible to make `protoc` available then we could include the build.rs in the project, although it shouldn't require changes in the definition again.
> 
> The case where we need to re-register (with exponential backoff) in case of failed communication with the hub will be added in the future.

Any ideas of how we can overcome this?

raftcat/Cargo.toml
```diff
@@ -33,3 +40,4 @@ chrono = "0.4.6"
 reqwest = { version = "0.10.2" }
 semver = { version = "0.9.0", features = ["serde"] }
 tokio = { version = "0.2.11", features = ["full"] }
+protobuf = { version = "2.16.2", features = ["with-serde"] }
```
clux, line 43: You are already enabling `with-serde` here.

raftcat/Cargo.toml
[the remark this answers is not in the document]
clux: It's already active as per line 43, this creates equivalent features on raftcat, which do nothing.

raftcat/Cargo.toml
carlosgaldino: This one only applies when generating the `.rs` from the `.proto`. It generates the:
`#[cfg_attr(feature = "with-serde", derive(Serialize, Deserialize))]` on top of each `struct` declaration. 

The one under `[features]` is used during the compilation to derive the traits, no?

I tried removing the `with-serde` from `[features]` and the compilation failed with a message saying the serde traits were not satisfied anymore.

clux: oh... ok. sorry, i didn't see the feature was used internally (only skimmed the generated diff)

raftcat/Cargo.toml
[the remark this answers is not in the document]
clux: nevermind, i'm wrong!

clux:
Protoc missing is due to it not being in the build image: https://github.com/clux/muslrust . You can raise an issue there to track, but not sure I want to add it there tbh.

That said, I wouldn't worry about the build image not auto-generating the protoc stuff all the time. It would still be nice to have the `build.rs` inside source so people could generate it locally if they wanted to.

raftcat/src/main.rs
```diff
@@ -254,6 +273,15 @@ async fn main() -> std::io::Result<()> {
     };
     let shared_state = state::init(cfg).await.unwrap();
 
+    let region = shared_state.get_region().await.unwrap();
+    let region_str = region.raftcat_url().expect("raftcat url");
```
carlosgaldino, line 277: @clux How does the [`base_urls` field in `Region`](https://github.com/babylonhealth/shipcat/blob/master/shipcat_definitions/src/region.rs#L589) gets populated? What sort of configuration should be set when running raftcat, so the `base_urls` are not empty?

clux: it's manually set in `shipcat.conf` for regions with a services entry point. it's not set everywhere, but think it's set everywhere raftcat is running. if you don't want to talk through raftcat through the services endpoint, or if it doesn't exist, you have to use cluster dns.

carlosgaldino:
@clux Thanks for your comments and review! I tried writing the `KOMPASS_AUTH_TOKEN` before merging this PR, but I don't have permission for it. What's the best way to proceed with this?

PR #476, opened 2020-08-14
DAT-455 extend es to include team metadata

nandaloussiBabylon:


shipcat_definitions/src/structs/eventstream.rs
```diff
@@ -23,6 +27,9 @@
 
     #[serde(default)]
     pub config: BTreeMap<String, String>,
+
+    #[serde(default = "default_metadata")]
```
clux, line 31: This is just the normal default. No need to define an extra fn.

shipcat_definitions/src/structs/eventstream.rs
```diff
@@ -8,6 +8,10 @@
     pub value: String,
 }
 
+fn default_metadata() -> BTreeMap<String, String> {
+    BTreeMap::new().into()
+}
```
clux, line 13: Remove this -_-

PR #477, opened 2020-08-17
[DXKP-120] Reject requests without authorisation

carlosgaldino:
This patch adds support for checking the authorisation header and reject
any requests made with an incorrect token.


raftcat/src/kompass.rs
```diff
@@ -12,7 +12,7 @@ use tokio::time;
 
 const REGISTER_PATH: &str = "register";
 const HEARTBEAT_PATH: &str = "heartbeat";
-const KOMPASS_AUTH_TOKEN: &str = "KOMPASS_AUTH_TOKEN";
+pub const KOMPASS_AUTH_TOKEN: &str = "KOMPASS_AUTH_TOKEN";
```
clux, line 15: not sure why you are defining a constant for this if it's only used once

carlosgaldino: It is also being used when sending the registration requests as well.

carlosgaldino:
@clux Thanks for the review. Could you please merge? I get `You’re not authorized to merge this pull request.`

clux:
Are you not in @dev-platform ? I added you guys to merge.

PR #497, opened 2020-12-01
Allow for an externally managed base chart

andyflatt:


sg-babylon (changes requested):
spelling in the docs but otherwise seems fine

shipcat_cli/src/helm.rs
```diff
@@ -36,6 +37,34 @@ pub async fn hout(args: Vec<String>) -> Result<(String, String, bool)> {
...
+pub async fn clone_chart(repo_url: &str) -> Result<(String, String, bool)> {
+    debug!("git clone {}", repo_url);
+    let path = format!("{}{}", "charts/", repo_url);
+    if Path::new(&path).exists() {
+        debug!("Repo already cloned {}", repo_url);
+        Ok(("".to_owned(), "".to_owned(), true))
+    } else {
+        let re = Regex::new(r"(^git@.*)\?ref=(.*)").unwrap();
+        if !re.is_match(repo_url) {
+            Ok((
```
gleaver, line 49: I assume you're intentionally return an `Ok` here rather than just bailing in order to give more useful error? You have an `await?` lower down that does just bail, though.

shipcat_cli/src/helm.rs
```diff
@@ -36,6 +37,34 @@ pub async fn hout(args: Vec<String>) -> Result<(String, String, bool)> {
     Ok((out, err, s.status.success()))
 }
 
+pub async fn clone_chart(repo_url: &str) -> Result<(String, String, bool)> {
+    debug!("git clone {}", repo_url);
+    let path = format!("{}{}", "charts/", repo_url);
```
gleaver, line 42: Can't work out if this is gross or elegant, but it will work.

gleaver (approved):
Not entirely clear to me that we want to allow arbitrary charts rather than a single versioned collection in the long run, since shipcat is partly tied to the chart and vice versa. This looks like a reasonable short-term option though.

Is this definitely preferable to just having multiple base chart versions in manifests for the upgrade? We'll need to upgrade the pins of every service (though we should be able to change those back if needed, and a full reconcile should result in a no-op if they're the same)

shipcat_cli/src/helm.rs
andyflatt: It is elegent

- Nice folder structure
- Can make edits whilst in the manifest repo and commit them back
- No need to pull more than once
- Zero code

shipcat_cli/src/helm.rs
andyflatt: Exactly as intended.

sg-babylon (approved):

PR #499, opened 2021-01-14
Bump some libs

hughsimpson:
- Fixes the ```thread 'main' panicked at 'attempted to leave type `linked_hash_map::Node<serde_yaml::value::Value, serde_yaml::value::Value>` uninitialized, which is invalid', /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/core/src/mem/mod.rs:658:9``` thing and seems to work with 1.49.0


gleaver:
Suggest changing the `cargo_test` build back to use `circleci/rust:latest`, and then running `scripts/bump_version.sh 0.159.0 0.159.1` to build a new version

gleaver:
Merged this into another PR

PR #500, opened 2021-01-14
[DEVP-1854] Pin toolchain to 1.49.0 in all CI images to avoid serde d…

gleaver:
…eserialisation error in stable.

Apply the toolchain changes the tests are using to both the build images as well, and upgrade to 1.45.2. Local testing seems to validate against actual manifests.

You can test this locally with
```
rustup toolchain install 1.45.2
cargo +1.45.2 build
```

Once approved/tested I'll roll a version pin change into this PR and make a new version


janslow (approved):

gleaver:
Added in Hugh's change to upgrade serde past the bug, and also pinned version to 1.49.0 to avoid future issues.

PR #510, opened 2021-03-12
[DEVP-1901] Formalise docs entries in metadata

gleaver:
Move our required docs fields out of custom metadata and into the proper metadata structure. MetadataSource is extracted in order to allow the parsing of the field as either a string or array of string in YAML. The CRD always holds these as a vector.


shipcat_definitions/src/structs/metadata.rs
```diff
@@ -193,12 +193,29 @@ pub struct Metadata {
...
+
+    /// Link to the Product Engineering Document for the service
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub ped: Option<String>,
+    /// Link to the test plan for this service
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub testPlan: Option<String>,
+    /// Link to the release plan for this service
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub releasePlan: Option<String>,
```
clux, line 205: `Option<Url>` should work and you'd avoid having to try parsing it

gleaver: This does indeed work, although the user ends up with a very opaque error from `shipcat validate` when its done as part of the deserialisation, so I've kept this implementation instead.

bh-anton-k (approved):

PR #519, opened 2021-06-29
[FLOW-1827] upgrade python version for circleci

crispin-clark:


Dockerfile.circleci
```diff
@@ -1,4 +1,4 @@
-FROM circleci/python:3.7.4-node
+FROM circleci/python:3.8.10-node
```
ryanb93, line 1: `circleci/python` has been deprecated in favour of `cimg/python` 

> This image is designed to supercede the legacy CircleCI Python image, circleci/python.

Should that be used instead while we're changing things?

ryanb93: Just a warning that it uses `pyenv` so the location of any hardcoded references to `site-packages` will have changed, got caught out here: https://github.com/babylonhealth/cd-workflows/pull/236/files#diff-a5f97da102a4ba5379ea1191b94f835106a75b8202903a1aef8391cd36a089b6R10

gleaver (approved):
Happy with this, though we might as well change the base tag as Ryan suggests

Dockerfile.circleci
crispin-clark: checked and there don't seem to be any in manifests either

PR #523, opened 2021-09-23
[CLOUD-2900] Ability to hash ConfigMap files in the ShipcatManifest CRD

dedoussis:
manifest.yml:
```yaml
configs:
  digest: true
  mount: /config/
  files:
  - name: config.staging.yml.j2
    dest: product_config.yml
  - name: logging.conf.j2
    dest: logging.conf
  - name: newrelic-python.ini.j2
    dest: newrelic.ini
```

ShipcatManifest CRD:
```yaml
  configs:
    mount: /config/
    files:
      - name: config.staging.yml.j2
        dest: product_config.yml
        value: 270BDC49091BDD30AEF4722C715FBE596941EE84688D98E869A31A7A6826CEB3
      - name: logging.conf.j2
        dest: logging.conf
        value: A0F695CCE2EA25516AC94957F73FE75EC1BC89C461B008AB06E0F364C51D0676
      - name: newrelic-python.ini.j2
        dest: newrelic.ini
        value: 53F6C573CF81C654909CE2C6B599F0280812E12F9804968AB0BEB001FFABD66E
    digest: true
```

If the `digest` field is omitted the original behaviour is preserved.


shipcat_definitions/src/crds.rs
```diff
@@ -17,7 +17,7 @@ impl From<Manifest> for ShipcatManifest {
         // but no secrets!
         assert_eq!(mf.state, ManifestState::Base);
         #[allow(clippy::redundant_clone)] // cloning the name is cheaper than cloning the manifest
-        ShipcatManifest::new(&mf.name.clone(), mf)
+        ShipcatManifest::new(&mf.name.clone(), mf.with_hashed_configs())
```
gleaver, line 20: I think this is complaining that there's no need to clone at all, since the function already owns the object

dedoussis: I cloned it because I don't want to mutate the existing `mf` object (in case it gets used elsewhere)

dedoussis: But it gets cloned all the time, even if the `configs` object does not include `digest: true` field. I could optimise it to clone only if there is a need for mutation.

shipcat_definitions/src/structs/configmap.rs
```diff
@@ -13,6 +14,8 @@ pub struct ConfigMap {
     pub mount: String,
     /// Files from the config map to mount at this mountpath
     pub files: Vec<ConfigMappedFile>,
+    /// Enable this switch to hash each of the config files before storing it in the CRD
+    pub digest: Option<bool>,
```
gleaver, line 18: Probably leave this as just a bool without the option?

shipcat_definitions/src/crds.rs
dedoussis: So you suggest there might be no need to `self.clone()` within `with_hashed_configs`?

shipcat_definitions/src/structs/configmap.rs
```diff
@@ -32,6 +35,20 @@ pub struct ConfigMappedFile {
     pub value: Option<String>,
 }
 
+impl ConfigMappedFile {
+    pub fn with_hashed_value(&self) -> ConfigMappedFile {
+        let mut hashed = self.clone();
+
+        if self.value.is_some() {
```
gleaver, line 42: I think 
```rust
if let Some(value) = self.value {
   ...
}
```

shipcat_definitions/src/manifest.rs
```diff
@@ -1214,6 +1214,17 @@ impl Manifest {
         }
         Ok(())
     }
+
+    pub fn with_hashed_configs(&self) -> Manifest {
+        let mut hashed = self.clone();
+
+        if self.configs.is_some() {
```
gleaver, line 1221: `if let Some ...`

shipcat_definitions/src/structs/prometheusalert.rs
```diff
@@ -44,8 +44,8 @@ pub struct PrometheusAlert {
 /// Alert severity enumeration.
 ///
 /// Represents the set of alert severities we allow in our Prometheus alerts.
-#[serde(rename_all = "lowercase")]
 #[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(rename_all = "lowercase")]
```
dedoussis, line 48: This was throwing an annoying compilation warning.

gleaver (approved):

PR #525, opened 2021-09-27
[CLOUD-2019] Hash configs within apply

dedoussis:
This change ensures that the same `Manifest` -> `ShipcatManifest` converter is used across the `apply` and the `crd` commands.

Note that the `ShipcatManifest::from` converter hashes the values of the configs (introduced in #523 ), to account for the resource size limitations.


janslow (commented):
Could you add a test case for this?

E.g., test that the output `shipcat crd fake-ask` is expected

janslow (approved):
nvm, this is covered by the tests added in https://github.com/babylonhealth/shipcat/pull/523

PR #528, opened 2021-12-07
[CLOUD-4100] Fix security vulnerabilities

dedoussis:
Addresses vulnerabilities with regards to Rust and Python dependencies.

* Updated the python deps included in the kubecat docker image (that is used in manifests), so that GitHub's dependabot does not complain anymore. There is a chance that this might break the existing CI of manifests, but I can proceed with another PR updating manifests to make it compatible with the latest versions.
* Did `cargo audit fix` which updated plenty of minor version pins.
* Based on the cargo audit report and the internal Blackduck report, had to do a major bump on the actix-web crate which is the web application framework used to serve raftcat. There are code changes in the raftcat package to accommodate this backwards incompatible bump.


raftcat/src/protos/registration_api.rs
```diff
@@ -6,7 +6,7 @@
 #![allow(clippy::all)]
 
 #![allow(unused_attributes)]
-#![rustfmt::skip]
+// #![rustfmt::skip]
```
dedoussis, line 9: Oddly, the protobuff source was never committed to version control. See relevant discussion: https://github.com/babylonhealth/shipcat/pull/475

Hence, I have to manually edit the autogenerated files 🤮

raftcat/src/protos/services.rs
```diff
@@ -6,7 +6,7 @@
 #![allow(clippy::all)]
 
 #![allow(unused_attributes)]
-#![rustfmt::skip]
+// #![rustfmt::skip]
```
dedoussis, line 9: No way to regenerate this piece of code. Had to manually edit the file.

raftcat/src/main.rs
```diff
@@ -29,77 +31,101 @@ use actix_web::{
 };
 
 // Route entrypoints
-async fn get_single_manifest(c: Data<State>, req: HttpRequest) -> Result<HttpResponse> {
+async fn get_single_manifest(c: Data<State>, req: HttpRequest) -> HttpResponse {
```
dedoussis, line 34: Latest version of actix-web is incompatible with the handlers returning a `Result< HttpResponse>` type.

raftcat/src/main.rs
```diff
@@ -29,77 +31,101 @@ use actix_web::{
...
 
-async fn get_kompass_hub_services(c: Data<State>, req: HttpRequest) -> Result<HttpResponse> {
+async fn get_kompass_hub_services(c: Data<State>, req: HttpRequest) -> HttpResponse {
     let req_token = req.headers().get("Authorization");
     if req_token.is_none() {
-        return Ok(HttpResponse::Unauthorized().finish());
+        return HttpResponse::Unauthorized().finish();
     }
-    let req_token = parse_basic_auth(req_token.expect("request token"))?;
+    let req_token = parse_basic_auth(req_token.expect("request token")).ok().unwrap();
```
gleaver, line 100: This will crash if the auth isn't basic, needs present but invalid case catching: https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default

raftcat/src/main.rs
```diff
@@ -124,18 +150,18 @@ fn parse_basic_auth(value: &HeaderValue) -> Result<String> {
     Ok(auth_token)
 }
 
-async fn get_service(c: Data<State>, req: HttpRequest) -> Result<HttpResponse> {
+async fn get_service(c: Data<State>, req: HttpRequest) -> HttpResponse {
     let name = req.match_info().get("name").unwrap();
-    let cfg = c.get_config().await?;
-    let region = c.get_region().await?;
+    let cfg = c.get_config().await.ok().unwrap();
+    let region = c.get_region().await.ok().unwrap();
```
gleaver, line 156: Got a lot of `unwrap()` in here :(. Might be cleaner to break out most of the func into another one which still returns `Result`, and only unpack that in `get_response`? That way we can keep all the `?` until we actually interact with actix

raftcat/src/main.rs
```diff
@@ -253,35 +282,39 @@ struct SimpleRegion {
...
 
-async fn index(c: Data<State>, _req: HttpRequest) -> Result<HttpResponse> {
+async fn index(c: Data<State>, _req: HttpRequest) -> HttpResponse {
     let mut ctx = tera::Context::new();
     let data = c
         .get_manifests()
-        .await?
+        .await
+        .ok()
+        .unwrap()
```
gleaver, line 291: ditto unwrap

PR #538, opened 2022-02-10
[CLOUD-4621] Validate cron schedules

palfrey:


shipcat_definitions/src/structs/cronjob.rs
```diff
@@ -55,3 +57,12 @@ pub struct CronJob {
     #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
     pub serviceAccountAnnotations: BTreeMap<String, String>,
 }
+
+impl CronJob {
+    pub fn verify(&self) -> Result<()> {
```
janslow, line 62: Rather than validating it, I think it should just be added as a parse-time check in `CronJobSource.build`.

https://github.com/babylonhealth/shipcat/blob/master/shipcat_filebacked/src/container/cronjob.rs#L37

palfrey: Done in 7eef3380ed607a3dd1a652bd1be5ddc5e04f38ff

palfrey: Just tested current manifests master with this and got the much more usable
```
shipcat::validate: Error: failed to build manifest for case-cards-backend in dev-internal
Caused by: Cron schedule parsing error: invalid value

shipcat: verify error: Invalid shipcat data in 1 files
```
out of it (which is absolutely correct and needs fixing, but I'm delaying on it until this is fixed)

janslow (approved):

PR #543, opened 2022-03-01
[CLOUD-4817] Update: Add non-root USER 1000 to raftcat; Run as USER 1000

cindywfeng:
This is part of security remediation work for the DivvyCloud insight: https://babylonhealth.customer.divvycloud.com/resources?insight=backoffice:365&resource_type=pod&active_category=container

The aim is to add a securityContext for the pods and set the runAsUser as a non-root user e.g. 1000


gleaver (approved):
Remember to bump the versions before merging

